PR #129, opened 2015-09-01
Update Dexie.js

jimmywarting:
Useful when you are dealing with promise... 
And its Better to resolve something then undefined

This could be reduced

	var db = new Dexie(obj.database);

	return db.open().then(function(){
		return db;
	});

To:

	return (new Dexie(obj.database)).open();


dfahlander:
Thanks

PR #130, opened 2015-09-02
Do not leak arguments

YuriSolovyov:
Prevent leaking of arguments according to [Optimization guide](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments)

### DIFF
--- a/src/Dexie.js
+++ b/src/Dexie.js
@@ -263,9 +263,10 @@
                                             }
                                         }
                                         return orig_promise.call(this, mode, function (resolve, reject, trans) {
-                                            arguments[0] = proxy(resolve);
-                                            arguments[1] = proxy(reject);
-                                            fn.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


YuriSolovyov:
sorry, I didn't get the `The arguments object must not be passed or leaked anywhere.` part.
to really not leak the args, you still need to do this inline.

YuriSolovyov:
@dfahlander do you want me to open a new PR with inline args copying?

dfahlander:
Not sure if it's worth it. I would suppose that type of optimization is
disappearingly small compared the time it takes to perform a db operation.
Maybe in some places (functions that are frequently called from within
loops) it could be.
Den 2 sep 2015 20:46 skrev "Yuri" <notifications@github.com>:

dfahlander:
In 1.4.0-beta, I've got rid of all leaking arguments.

PR #133, opened 2015-09-09
Added Dexie.exist

jimmywarting:
Usage: 

    databaseExists("foo").then(AlreadyTaken, createDatabase)

Not sure if its a good idea to use reject like i do.

Maybe should be:

    databaseExists("foo").then(function(result){
       result ? AlreadyTaken() : createDatabase()
    }, function(err){
       console.log(err)
    })

### DIFF
--- a/src/Dexie.js
+++ b/src/Dexie.js
@@ -3264,6 +3264,42 @@
         return fn && fn.bind(indexedDB);
     }
 
+    //
+    // Static delete() method.
+    //
+    Dexie.exists = (function(){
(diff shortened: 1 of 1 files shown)


dfahlander:
Wrote a unit test for it, and yes, I think it's better to always resolve with true/false because it's not an error if it doesnt exist. It was possible to simplify the method by testing to open it with Dexie without specifying a schema. This will result in an error if the database didnt exist. See commit 5229ea5523e69ec98495ce2123fc15346c1b975f

PR #138, opened 2015-09-22
Performance boost

jimmywarting:


dfahlander:
Thanks. Need to fix slice() though because it has to be used with apply() or bind().
Would you have a chance to run the unit tests in Chrome and Mozilla? Just open dexie/test/run-unit-tests.html in you browsers and wait for them to complete.

Thanks again,
David

jimmywarting:
601 assertions of 601 passed, 0 failed. both chrome and firefox. don't see any problem with apply or bind...?

dfahlander:
Ok, I must have looked at it too fast. missed out that you already wrapped _slice with a function that did _slice.call().

dfahlander:
I put you add a contributor, so if you find more nice things to improve, please go ahead and commit directly. I can review before releasing to dist and npm.

jimmywarting:
Sweet, thanks!

PR #147, opened 2015-10-28
Allow dexie to be referenced in typescript projects.

isuda:
e.g.

```
import * as Dexie from 'dexie';
```

Dexie will now be properly typed without having to reference Dexie.d.ts in DefinitelyTyped or through `src/Dexie.d.ts` (which isn't exposed in npm anyways)

If you don't want to duplicate Dexie.d.ts then you could also change the typings line to point to `src/Dexie.d.ts` but you would also need to add an exception to `.npmignore`


dfahlander:
Sounds good. Didn't know that was possible. Does that mean that Definitelytyped and tsd aren't needed and I could delete Dexie from Definitelytyped? Does it work the same in various IDEs and Typescript versions?

isuda:
This should work with Typescript >= 1.6. It is working for me in Atom (atom-typescript plugin) and in WebStorm. I haven't tried other IDEs.

https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages

isuda:
Yeah, if you use the module this way, you won't need definitely typed typings anymore. But I would probably leave them there as well for people that prefer to do it that way. :)

dfahlander:
Note: Need to update the typescript samples for this change + make them module based and TS1.6 compatible.

samuelhorwitz:
I'm pointing my NPM to the master branch of this repo to test this out and I see that when I do the following, the build seems to be fine but the code itself fails:

```typescript
import * as Dexie from 'dexie';
export class MyDB extends Dexie {
...
    constructor() {
        super('MyDB');
    }
...
```

Like I said Typescript likes it just fine but actually loading the page leads to an error about calling `super` due to it being an object not a function. Looking at what `Dexie` actually is upon export, it is in fact an object (with a key `default` that looks like what I want). For some reason, there appears to be a discrepancy between what the typings say and what the actual code has, although I'm fairly new to Typescript so I'm not sure if I'm running into something else.

I'm running Browserify, Typescript 1.6 -> ES6, Babelify.

isuda:
@samuelhorwitz: `__esModule` sounds like something that typescript adds. 

When it comes down to it, I am doing it the same way:

```
import * as Dexie from 'dexie';
export class AppDatabase extends Dexie {

  static databaseName = "AppDatabase";
  public $injector: ng.auto.IInjectorService;
  private initialized = false;
  private initializedCallbacks: Array<Function>;

  static migrations: {
    [version: number]: Array<{name: string, migration: IDBMigration}>
  } = Object.create(null);

  constructor() {
    super(AppDatabase.databaseName);
  }

**Comment by isuda** [comment]
My workflow is using Browserify, Typescript -> ES5.

**Comment by samuelhorwitz** [comment]
It's definitely something to do with Typescript to ES6. I'll try and look into it a bit more, my solution for just ignoring it for now is really bad, but involves adding that super value of `Dexie.__esModule = true` right below the version setter in the `node_modules` Dexie file. Obviously this is not actually a solution but a stopgap.

**Comment by samuelhorwitz** [comment]
Hey guys quick update:

I have a *slightly* and I mean **slightly** better hack for this right now.

In my Gulpfile I define the following transform on my Browserify instance:

```coffeescript
b.transform (file) ->
    through (buf, enc, next) ->
        key = null

        if (_.find forceEs6FileNames, (f) -> key = f ; !!file.match f)
            buf = handleModule buf, key

        @push buf
        next()
        return
, global: true
```

This calls a function:

```coffeescript:
handleModule = (buf, file) ->
    str = buf.toString 'utf8'
    str = str.replace new RegExp("#{forceEs6ModuleMap[file].lineMatcher}.*"), (str) -> "#{str}\n#{forceEs6ModuleMap[file].global}.__esModule = true;\n"
    return new Buffer str
```

And at the top of the file, I need to declare some mappings (in this case, just for Dexie):

```coffeescript
forceEs6ModuleMap =
    'Dexie.js':
        lineMatcher: 'Dexie.version'
        global: 'Dexie'

forceEs6FileNames = _.map forceEs6ModuleMap, (v, k) -> k
```

Anyway, the better solution here is to support ES6 modules natively I suppose or maybe there is something that does this and I just rewrote it :D Anyway it's kind of painful to watch these ecosystems become more and more of a mess (nothing against the Dexie team, more a comment on JS in general). Hopefully it's just the rough patch before the golden dawn of native browser modules "just working".

PR #166, opened 2016-01-17
Don't try to modify frozen objects.

dpogue:
If an immutable object is passed to `put` or `add`, trying to assign to the keypath will throw an exception. In that case, we should just not attempt to modify the object.


dfahlander:
Thanks!

PR #174, opened 2016-01-31
Enable testing with karma

YuriSolovyov:
This adds ability to test Dexie.js from command line and opens a possibility of spinning up a CI.

For now, I only added core tests, addons test are hader to add because some of them are hardcoded into html files and we need to extract and maybe move them to `tests/addons` folder, so that `addons/` folder will only contain sources and dists.

I've chosen `mocha` reporter due to its readability, but we can discuss that.

Note that:
```javascript
"engines": {
   "node": ">=4.2 <5.0"
 },
```
is relevant only for potential CI env, not real platform support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,3 +160,5 @@ samples/typescript/console.js
 samples/typescript/console.js.map
 samples/typescript/utils.js
 samples/typescript/utils.js.map
+
+node_modules/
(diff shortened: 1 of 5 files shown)


dfahlander:
Thanks a lot, @YuriSolovyov. Merged it locally, installed and tested. Works fine. Added qunitjs to devDependencies since there was a warning about it missing.

Issue #175, opened 2016-02-09
Dexie.js fails to store objects when the primary key is 0


bennyn:
Hello, I am using Dexie.js in my project with out-of-line keys, so that I need to specify the primary key on my own when I want to store an object in my database. This works pretty good but I encountered the issue that Dexie.js fails to store objects when the key is 0 (the number zero). 

It work's fine if I stringify it (like "0") but with a plain 0 it will respond with the message: `There was an error: Failed to execute 'put' on 'IDBObjectStore': The object store uses out-of-line keys and has no key generator and the key parameter was not provided.`.

I realized it when I switched from [ydn-db](https://github.com/yathit/ydn-db) to Dexie because ydn can handle zeros.

Here is the code I used to reproduce the problem:

```javascript
var db = new Dexie('myDatabase');
db.version(1).stores({
  myStore: ''
});

db.open()
  .then(function () {
    return db['myStore'].put({name: 'three'}, 3);  // works
  })
  .then(function () {
    return db['myStore'].put({name: 'two'}, 2);  // works
  })
  .then(function () {
    return db['myStore'].put({name: 'one'}, 1);  // works
  })
  .then(function () {
    return db['myStore'].put({name: 'zero'}, 0);  // breaks
  })
  .catch(function (error) {
    console.error('There was an error: ' + error.message);
  });
```

721687:
我开发chrome扩展，使用 put 这个方法时，开发版本的代码运行没有问题，但打包后的代码使用次方法报总是报错。后来研究了半天，发现是因为  `db.version(1).stores({
  myStore: ''
});`  这里面打代码变了，而我并未将 version(1) 改成 version(2)，导致使用最早那个版本的人运行代码就会报错。 也可以清浏览器端的保存的所有indexedDB数据，就不会出错了。

PR #176, opened 2016-02-09
Supporting integers as primary keys, resolves dfahlander/Dexie.js#175

bennyn:


dfahlander:
Thanks a lot for that. I had to add support for empty keys as well.

Added a unit test in tests-tabel.js.

PR #203, opened 2016-03-21
Fix typo in database name in Typescript example in Readme

ChrisPearce:
Fix database name type in TypeScript example in Readme


dfahlander:
Thanks

PR #285, opened 2016-07-02
Confirm table deletion

chrahunt:
It looks like Dexie doesn't remove the underlying IDB object store on upgrade unless an upgrader function runs. I updated the relevant test so it correctly identified that `store1` was still present in the backing IDB and updated the check that was causing the deletion to not occur.

### DIFF
--- a/src/Dexie.js
+++ b/src/Dexie.js
@@ -300,7 +300,7 @@ export default function Dexie(dbName, options) {
                 }
             });
             queue.push(function (idbtrans) {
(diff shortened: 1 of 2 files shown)


dfahlander:
Thanks! Don't know why the build fails. I've configured browserstack env vars in travis. Trying to restart the build...

dfahlander:
Seems travis doesnt invoke encrypted or configured env variables when triggered from a pull request for securíty reasons. Trying to fix this by using local firefox to launch tests on pull requests henceforth.

dfahlander:
Future PRs will run the test-suite using travis' firefox instead of Browserstack.

Issue #273, opened 2016-06-21
Problem with opening exsiting database


michael2043:
I worked with the code below:
a.html
var db = new Dexie("idb01");
db.version(1).stores({ tbl03: "[id1+id2],name,age" });
db.open();
.......
other operation...
........

when I use 'document.location.href="b.html"' and in b.html:
var db = new Dexie("idb01");
db.open();
db.tbl03.xxxx(other operations...)
"db.tbl03" is undefined

how can I use exsiting database in another page.

dfahlander:
The best thing is of you define your DB in a common file, for example db.js and then include (or require) that js file  from both a.html and b.html.

dfahlander:
Since I get this question quite often, I think it deserves an explanation to use somewhere in the wiki as well.

There are two modes that you can open a database in with Dexie:

1. Open with version and schema.
2. Open without version or schema.

## Open with version and schema

This is the most common use case, as well as the preferred one. You have a JS module where you define your database schema. You include that common js file from all your apps / html pages that should use that database. The first HTML page that run will create the database. The second time db is run, it will be used and the schema will serve as an API populator and set the direct table methods onto your db.

db.js
```js
var Dexie = require('dexie'); // (if using modules. Otherwise skip this line)

var db = new Dexie('dbname');
db.version(1).stores({
    // schema goes here
});
db.open().catch(function (err) {
    console.error ("Fatal: Failed to open database. " + (err.stack || err));
});
export = db; // If using modules. Otherwise skip this line.
```

app1:
```js
var db = require('./db'); // (if using modules)

// Use db here

```

app2:
```js
var db = require('./db'); // (if using modules)

// Use db here too.

```

## Open without version or schema

Dexie will read the installed database and provide an API to use it, but you wont get the implicit table properties directly at your db instance. To use a table, you must go via db.table('tableName').method(). The downside with this way is that you make your code dependent on something that you cannot control. What happens if the user clears it's database? Your code will stop working wherever you expect a table to exists. You would need to add lots of error checkings if so.

```js
new Dexie('dbname').open().then(function(db) {
    var tbl03 = db.table('tbl03');
    tbl03.doSomething()...
}).catch ('InvalidTable', function (err) {
    // This error may happen if your user has cleared out the database.
});
```

michael2043:
"Open with version and schema" means every time the page reload ,the "new Dexie()","db.stores()","db.open()" will be fired.

dfahlander:
[...]

PR #290, opened 2016-07-11
Delete table only if new schema explicitly sets it to null

prannayb:
The double equal also returns true if a value is undefined. Meaning, if the new schema does not have a table name, we'll delete it. This breaks the "updates-only schema for new versions" pattern recommended by dexie best practices. To fix this, explicitly check for null.

### DIFF
--- a/src/Dexie.js
+++ b/src/Dexie.js
@@ -386,7 +386,7 @@ export default function Dexie(dbName, options) {
     function deleteRemovedTables(newSchema, idbtrans) {
         for (var i = 0; i < idbtrans.db.objectStoreNames.length; ++i) {
             var storeName = idbtrans.db.objectStoreNames[i];
-            if (newSchema[storeName] == null) {
+            if (newSchema[storeName] === null) {
                 idbtrans.db.deleteObjectStore(storeName);
             }
         }


dfahlander:
Thanks for the fix. Currently, @chrahunt is rewriting the upgrade handling so I leave it to him to decide if this pr can be merged or not. Seems a unit test fails for upgrade though which would need to be investigated.

chrahunt:
Hi @prannayb, were you actually having an issue with tables being deleted when they shouldn't? There are several tests covering this case that are currently passing. [This](https://github.com/dfahlander/Dexie.js/blob/master/test/tests-upgrading.js#L139) is one example. It is important to note that the new schema is derived from all previous versions, so this line should not be causing trouble unless you have not specified a previous version (a currently unsupported configuration as described [here](https://github.com/dfahlander/Dexie.js/issues/273#issuecomment-227404692)).

On the basis of that I am going to close this PR, but please make an issue if you are experiencing any problems.

prannayb:
Hey Chris, I was only trying to upgade two versions changes at once. I had a need to change the primary key of a table. So I created two versions upgrades, v2 had table marked as null and v3 had table with new primary key. I didn't want to preserve table data, but I couldn't get dexie to allow me to do this. My PR was only something I thought was a bug that it would delete tables not in the currently-being-processed version upgrade. I'll wait on the primary key upgrade that you're working on instead. Thanks for the review!

Issue #287, opened 2016-07-03
Upgraders have a false API towards their object stores


dfahlander:
Related to #105 and #1. When an upgrader function attached to a Version instance accesses the provided upgrade transaction (trans), it will only have the object stores from the latest version in it, and believe that indexes from latest versions apply in the upgrader. In practice, you can never delete an object store (setting it to null in a later version) and having an upgrader function reading from it. Secondary problem: If a previous version had an index that has been deleted in a later version, an upgrader cannot utilize the previous index for upgrading to the newer version.

This needs to be fixed and it needs an architectural change.

Todo:
- Let each Version instance have a Transaction instance populated with the object stores that are relevant in that particular version, as well as the available indexes. Should be based on the previous version's stores and indexes plus added stores in current version.
- Stop populating Transaction.prototype with tables. Instead populate the prototype of a Transaction subclass bound to the latest Version instance. Create such Transaction instead of the base transaction on normal transactions. On upgrade transactions, create a transaction based on the version being upgraded.

Also, if possible make upgrader fail if trying to use db.table-style instead of trans.table.

chrahunt:
Do we still want to make upgrader fail on db[table] style access?

dfahlander:
We could allow db[table] style access. I close.

dfahlander:
The issue was solved 2018-06-03 in 3.0.0-alpha.3 with PR #714. Don't understand why I closed it in 2017 already... Anyway, it's solved now.

Issue #105, opened 2015-05-31
Error trying to apply upgrade function and object store deletion in same version


chrahunt:
This may be related to #1, but I think it's narrow enough in scope to warrant a different issue. This is in a Chrome extension so the IE bug is not relevant.

I have an existing database that I'm trying to use with Dexie. In the new version I need to remove a few data stores and run an upgrade function. The data from the data stores to be removed is used in the upgrade function. Based on the order that functions are applied in the version upgrade queue defined in [this](https://github.com/dfahlander/Dexie.js/blob/666f641cd6786a0ad3fe7b9879973ead527ac99f/src/Dexie.js#L182) function, it looks like the object stores are going to be deleted after the upgrade function is applied, so this shouldn't be a problem.

Existing object stores can be removed by setting the key name to `null` in the object passed to the `stores` instance method, but when I do that I receive an error on [this](https://github.com/dfahlander/Dexie.js/blob/666f641cd6786a0ad3fe7b9879973ead527ac99f/src/Dexie.js#L394) line because the tableSchemas for the transaction are only being derived from the current version, which now omits the previous store schemas.

I have a failing test on [this](https://github.com/chrahunt/Dexie.js/tree/delete-and-upgrade) branch. I can work on a fix later today, but I wanted to make sure I'm not missing some detail that would be necessary to use these to-be-deleted object stores in the upgrade function.

dfahlander:
Thanks for detailed explanation! I'm sure this is a bug indeed. Deep in my mind I recall that this is an unsupported and untested use case. Would be great if you did find a solution for it, please go ahead. And thanks muc!

dfahlander:
I saw you have a fix in that branch. Please do a pull request when you feel comfortable with it.

I tried to run the [test suite](https://rawgit.com/chrahunt/Dexie.js/delete-and-upgrade/test/run-unit-tests.html). It worked for all tests but [this one](https://rawgit.com/chrahunt/Dexie.js/delete-and-upgrade/test/run-unit-tests.html?testNumber=76). Seems to be a trivial issue that I could look at if you do a pull request.

chrahunt:
Working on this again after rebasing on master. I think there's a little more to this that isn't currently being tested (ensuring only applicable object store names are available on transaction at upgrade time, handling arbitrary version definition ordering) which I can also look at.

dfahlander:
[...]

PR #293, opened 2016-07-14
Upgrade fixes [in progress]

chrahunt:
As discussed, here is a start towards some of the upgrade-related changes necessary to address #1, #105, and #287.

Still to do:
- [ ] There may be some bug relating to table names being set on transactions or the final Dexie instance.
- [ ] Test with Syncable/Observable
- [ ] More thorough tests for upgrading/moving data from table to table during upgrades.

### DIFF
--- a/test/tests-upgrading.js
+++ b/test/tests-upgrading.js
@@ -1,12 +1,20 @@
 ﻿import Dexie from 'dexie';
-import {module, stop, start, asyncTest, equal, ok} from 'QUnit';
+import {module, test, equal, ok, deepEqual} from 'QUnit';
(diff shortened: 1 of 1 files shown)


chrahunt:
Updated my branch after rebasing on master, going to proceed using the approach discussed in #105.

dfahlander:
Thanks!

dfahlander:
Tests failed browserstack Ie11 with that an object store wasn't deleted as expected. I suppose it's due to the special treatment of IE in respect to its bug of object store deletion after read. (Travis runs the unit tests on FF only when not having the browserstack credentials as env vars).

Issue #356, opened 2016-10-24
Dexie.d.ts needs ReadonlyArray as an option of IndexableType


wwoods:
As title.  Basically, if you have a `ReadonlyArray`, it currently won't work with e.g. `anyOf()` because that function only takes an `Array`, not a `ReadonlyArray`.  Naturally though, `anyOf()` doesn't require a mutable array.

wwoods:
Also: get() should specify whether it returns `null` or `undefined` when the object does not exist, in conformance with `--strictNullChecks` in Typescript 2.

dfahlander:
Thanks!! If I forget to follow up on this, please send a pull request with the fix (should commit the change to src/Dexie.d.ts) or remind me ;)

wwoods:
Minor update - `IndexableType` needs to cover all bases with respect to Dexie.maxKey (needs to include `string[]` looks like):

    Argument of type '(string[] | Done)[]' is not assignable to parameter of type 'IndexableType'. Type
            '(string[] | Done)[]' is not assignable to type '(string | number | Date)[]'. Type 'string[] | Done' is not
            assignable to type 'string | number | Date'. Type 'string[]' is not assignable to type 'string | number |
            Date'. Type 'string[]' is not assignable to type 'Date'. Property 'toDateString' is missing in type
            'string[]'.

wwoods:
declare type IndexableTypeParts = string | number | Date | null | Array<Array<void>>;
    declare type IndexableType = IndexableTypeParts | Array<IndexableTypeParts>;

    static maxKey: Array<Array<void>> | string;

It's on my list to make a pull request with this stuff.  Should also add `static minKey: number`... and set it to `Number.NEGATIVE_INFINITY`: https://www.w3.org/TR/IndexedDB/#dfn-key.

dfahlander:
Thanks for clarifying these things out, and also for planning to create a PR. Much appreciated :+1:

wwoods:
Awesome, thanks for the merge :)

PR #371, opened 2016-11-22
Issue 356

wwoods:
Two commits to fix Issue #356 - 

1. A commit adding `minKey`
2. A commit adding `undefined` where appropriate, and specifying `ReadonlyArray` as an option for others.


dfahlander:
Thanks a lot for this work. :+1:

Issue #374, opened 2016-11-22
Calling WebCrypto API from within an upgrade transaction


dfahlander:
Continuing the question from @sechel on #317 into this own issue.

### Background

WebCrypto is a promise based API that does not play well with IndexedDB transactions on any browser due to the characteristics of IndexedDB transactions where all async API's will make a transaction fire off except if the API called was against the transaction itself. The reason WebCrypto is asynchronic, is that it enables the API to do heavy computations within a worker thread instead of the GUI thread.

This leaves 2 options for encrypting data with IndexedDB:

1. Use another JS based synchronous crypto library instead of WebCrypto. Not all encryption algorithms are CPU intensive and many of them wouldn't need to be executed by worker threads-could be to be async anyway except for some performance gain on some heavy cryptographic computations. If using symmetric key encryption, it's pretty fast anyway and can be done in the same thread.
2. Use WebCrypto, but prepare the encryption before starting an indexedDB transaction.

### Upgrading Solution

@sechel needs to encrypt data when upgrading from version A to version B. He wants to use WebCrypto for encrypting the data, but it is impossible to invoke withing an upgrade transaction so he needs another solution.

A working solution would need to combine an upgrader function with the [db.on('ready')](https://github.com/dfahlander/Dexie.js/wiki/Dexie.on.ready) event. The upgrader can flag entities for being in need of encryption, while db.on('ready') listener can call WebCrypto and perform the actual encryption.

[db.on('ready')](https://github.com/dfahlander/Dexie.js/wiki/Dexie.on.ready) will be called whenever the indexedDB.open()'s onsuccess event has been triggered, but before Dexie's db.open() promise resolves. From there, it is possible to do additional asynchronic preparation of the database before the it will be considered open.

By returning a promise from your event listener of db.on('ready'), the db will be blocked )for all but yourself) until your promise resolves so that your app code don't need to think waiting for your upgrader to complete.

Here's my suggestion:
```js

db.version(A).stores({myTable: 'id, indexes...'});
db.version(B).stores({myTable: 'id, indexes..., isEncrypted'}).upgrade(()=>{
    // Mark all entities to be unencrypted:
    return db.myTable.toCollection().modify({isEncrypted: 0});
});

db.on('ready', encryptIfNeeded);

const LIMIT = 1000; // Have some kind of limit to spare memory and be able to show progress...

function encryptIfNeeded () {
[...]

PR #378, opened 2016-11-24
New tool: Dexie.waitFor() keeps transaction alive.

dfahlander:
Inspired by @wwoods solution described  in #374 and the new [indexeddb-promises proposal](https://github.com/inexorabletash/indexeddb-promises/issues/11), I've added the method Dexie.waitFor() that may wait on a non-indexedDB Promise or Promise-returning function to complete and keep transaction alive during the whole lifetime of that task.

Implementation was also inspired by the elegant code in [polyfill.js](https://github.com/inexorabletash/indexeddb-promises/blob/master/polyfill.js) by @inexorabletash.

### DIFF
--- a/src/Dexie.js
+++ b/src/Dexie.js
@@ -1387,13 +1387,13 @@ export default function Dexie(dbName, options) {
         this.on = Events(this, "complete", "error", "abort");
         this.parent = parent || null;
         this.active = true;
-        this._tables = null;
         this._reculock = 0;
         this._blockedFuncs = [];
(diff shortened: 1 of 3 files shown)


sechel:
Typescript type definition update is missing.

wwoods:
@dfahlander Looks good to me, and should allow cleaner code than the writer / reader "threads" I was using.  Perhaps a smidge less performant, but more than made up for by ease of use.  Nice work :+1:

PR #395, opened 2016-12-12
Refactoring upgrade process

### LINKED ISSUE #105: Error trying to apply upgrade function and object store deletion in same version

chrahunt:
This implements the upgrade process described in #105. Instead of opening the database at the latest version and artificially moving through each version (while still under the same transaction), we now open the database at each version in turn. At each upgrade we diff between the current state of the db and the version schema applicable to the current version of the db. This should have the following effects:

1. Those mentioned in #105
2. Remove the need to keep around old versions of object stores if an upgrade is not necessary
This also removes the need to keep around old versions of object stores if there is not an upgrade necessary from version to version. This assumes you 'bring forward' any implied table definitions from previous versions to be defined on the current one.

This should resolve #105 as it makes the correct object stores available on the upgrade transaction. Some discussion is needed for whether this brings us closer to #287.
Additionally, this adds a 'max version' parameter on explicit DB open that can be used for more controlled upgrade and easier unit testing.

## Summary of changes:

1. Several methods were moved to be static methods of the associated class type, e.g. `TableSpec` now has static methods for parsing from Version store source and from a db or transaction. Likewise `IndexSpec` has a static method for parsing from the Version store source syntax.
2. Version static methods added for working with version numbers and versions that have been set.
3. `runUpgraders` function responsibility split into `doUpgrade` (called when we are upgrading an existing database) and `create` (called on initial db creation, so `on.populate` is called).
4. `updateTableAndIndexes` (now named to `upgrade`) is called from `doUpgrade` to do the actual upgrade. Anything regarding looping over versions has been removed since we're only upgrading one version at a time.
5. `this._open added` as the simplified version of `this.open`, called from within `this.open`
6. `this.open` now opens the database without a version and either creates it initially if `onupgradeneeded` fires or closes the database and delegates to `this._open` if a newer version is available.
7. `max_version` parameter added to `this.open` for ease of unit testing upgrades
8. `Transaction` was changed so its list of store names can be updated.
9. We still stick to setting `Table` instances on `Transaction.prototype`, `db`, and `allTables` to make them available. I tried to make it more explicit exactly where we were setting these and from where we were deriving the schemas.

## Pending activities

1. [x] There is one failing test in `WhereClause#anyOf` regarding Classes being defined on the Tables. Not sure the best way to resolve that. I started working on a class cache for Tables that could be re-set on `Table` initialization but I don't know if this is relevant to point 3 on your comment [here](https://github.com/dfahlander/Dexie.js/issues/105#issuecomment-262737335).
2. [x] Need to remove `readGlobalSchema` as it is unused.
3. [x] ~~We may want to have some validation between versions that happens when versions are added as opposed to on upgrade.~~ - will leave this to another issue, there's enough here already
4. [x] Update needed to upgrading tests to cover claims that we don't need previous version stores specified (these could also be made more thorough in general but that could be saved for another time).
5. [x] Remove `for ... of`
6. [x] Use `extend` instead of `Object.assign`
7. [x] Address destructuring
8. [ ] Fix `dynamicallyOpened` to reflect whether at least one version has been specified

## Open items

1. `createMissingTables` in `upgrade` - I'm not familiar with Dexie addons that use this, if it could be addressed that would be good.
2. `IDBObjectStores` dependency


src/Dexie.js
```diff
@@ -472,92 +529,180 @@ export default function Dexie(dbName, options) {
...
+                let idbdb = e.target.result;
+                let dexie_version = Version.dbToDexie(idbdb.version);
+                let next_version = Version.getNext(dexie_version, max_version);
+                if (next_version) {
+                    // Close database and then open at next version.
+                    db._close(idbdb);
+                    resolve(db._open(next_version, max_version));
+                } else if (dexie_version > max_version) {
+                    throw new exceptions.OpenFailed(
+                        `Max desired version ${max_version} is higher than` +
```
chrahunt: Should say `lower`

chrahunt:
1. Object.assign: Is it alright to move to a babel plugin instead? [object-assign-transform](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-object-assign)
2. destructuring change: :+1:
3. for .. of: Do you want the usages within this PR reverted or another issue opened to track a move to bublé?
4. Is [this](https://github.com/dfahlander/Dexie.js/pull/395/files#diff-1895b29b9e9bc7e2abef8890f993e5b2R3455) not the appropriate way to ensure we're covered in the idbshim case?
5. Hooks: I will look into this today or tomorrow.

dfahlander:
1. Use extend() instead (imported from './utils'). Lesser output size. The _extend helper would be put in every module instead of just used once. If needed we could extend "extend" to allow multiple arguments.
2. 👍 
3. Revert to not using for..of so far. Going over to buble is a bigger task even though I'd really like to test it out. If going over to buble to build dexie, we might need to keep babe as well to build the unit tests since we're using async/await in the tests, which are not supported by buble. *[dexie-relationships](https://github.com/ignasbernotas/dexie-relationships) addon successfully moved over to buble/rollup for the build. Could be used as a reference.*
4. In older iOS versions, the shim is unable to be applied to window.indexedDB due to the iOS WebView restrictions, so people need to use the shim by setting Dexie.dependencies.indexedDB and Dexie.dependenciea.IDBKeyRange, OR they can provide them as options to Dexie constructor.
5. <3

dfahlander:
I merged a minor change that resulted in a conflict for this PR. I can fix if you want (when it's time to merge). The conflict is somewhere in db.open() where globalDatabaseList() was called on so I think it can be easily resolved.

nponiros:
Adding this here as a note so we don't forget it.

From what I can tell the \_parseStoreSpec() will be removed by this PR. We would have to adjust Dexie.Observable which uses the method to define its stores. I don't know if any other changes affect the addons.

chrahunt:
This is ready for review and (potential) merge, pending a look at `createMissingTables` by someone familiar with the addons that use it, and confirmation that we can add `IDBObjectStore` to dependencies for release 2.

nponiros:
I have been trying for a few hours now to get Dexie.Observable and Dexie.Syncable to work with the PR with not much luck. There seem to be a number of issues we have to deal with and I don't know what the best course of action is.

First off this PR changes dynamicallyOpened() to always return `true`. Dexie.Observable uses this function to decide, among others, when it needs to observe tables. With dynamicallyOpened() always returning `true` it never observes any table which pretty much breaks everything. What is the reason for it to always return `true`? It used to return `false` if at least one DB version was defined.

Second Dexie.Observable adds new stores to the database. I managed to get that to work by overriding `Version.prototype.stores` and adding the needed stores. This seems to work without an issue. (Used to be in the _parseStoreSpec() override)

Third Dexie.Observable also changes the schema of the tables in two ways:
* It adds `schema.observable = true` so we know which tables have to be observed. I managed to get that to work by overriding `Dexie.TableSchema.fromVersionStores` and `Dexie.TableSchema.fromTransaction` and add the code to set the `observable` property (used to be in _parseStoreSpec() override).
* It defines a new index type `$$` for the primary key to create unique IDs instead of using `++` or user defined IDs. In order to do that it reads `schema.primKey.name` and checks if `$$` is at the beginning of the name. The problem here is that this PR seems to change the name attribute (probably [line 3178](https://github.com/dfahlander/Dexie.js/pull/395/files#diff-1895b29b9e9bc7e2abef8890f993e5b2R3178) and that breaks the logic of Dexie.Observable causing Dexie to think that we are changing the primary key which causes it to throw `"No support yet for changing primary key"`.
This is the logic in Dexie.Observable used for `$$` (used to be in _parseStoreSpec() override, is now in fromVersionStores() and fromTransaction() overrides):

```
if (schema.primKey.name.indexOf('$$') === 0) {
  schema.primKey.uuid = true;
  schema.primKey.name = schema.primKey.src.substr(2);
  schema.primKey.keyPath = schema.primKey.src.substr(2);
}
```

Using only fromVersionStores() or fromTransaction() also breaks things because as it seems the schema changes made by Dexie.Observable are not persisted. Maybe the easiest solution for this would be to add `$$` to Dexie directly and let it create the IDs itself. From what I can tell adding this to Dexie should increase its size by much.

I also had to add `globalSchema = db._dbSchema = this._cfg.dbschema;` below [line 209](https://github.com/dfahlander/Dexie.js/pull/395/files#diff-1895b29b9e9bc7e2abef8890f993e5b2R209) since without it _createTransaction() was sometimes called without a schema meaning Dexie.Observable couldn't read the `observable` attribute and that broke some logic in Dexie.Observable.

It would be nice if we find solutions for these issues before this PR is merged so that Dexie.Observable/Dexie.Syncable don't break and so that we can provide an upgrade path for any addons adding tables and/or changing the schema of the database.

As a side note [dexie-relationships](https://github.com/ignasbernotas/dexie-relationships) also uses _parseStoreSpec() and also changes the schema. So it will also have to be upgraded to work with this PR.

chrahunt:
1. `dynamicallyOpened` - My fault, I interpreted this to indicate that we're deriving the schemas from the database itself, and that is always true with this change. I don't see a problem fixing it to reflect 'at least one version is defined' if that's what it is meant to be.
2. The rest of the issues seem like a problem with hooking into Dexie, in most cases for changing database specs. Can we create a more consistent and known public interface to deal with these cases? Ideally, we should never have to worry about external add ons when changing internal behavior (especially if we can't even trust `_`-prefixed methods). From what you've stated so far we need:
  1. Method for adding custom index types on user tables
  2. Method for adding custom tables
  3. Hook into transaction creation

nponiros:
1. Would be cool if we can just revert that change to do what it used to do.
2. Yes a more consistent public interface would be good. The main issue as you noted is that the addons use `_`-prefixed methods which should probably be considered private.
  i. Yes
 ii. I think hooking into `stores()` is good enough for that as `stores()` is public (overrides are used for other public methods as well like open() and close())
iii. Yes. Although currently (for Dexie.Observable) it is good enough when _createTransaction() always gets the schema
vi. A way to change the schema of tables independent of any custom index (could of course be only one method which allows schema and index changes)

dfahlander:
Thanks guys for identifying these issues <3  I hope to get some time during the coming week to dive into these things. Agree we should have a non private api for the identified overrides

chrahunt:
I will fix `dynamicallyOpened` in this PR. Is it alright to open an issue for 2 (defining new API to cope with breaking changes to internal methods) and address it separately? I would prefer this PR not be blocked on it.

dfahlander:
Absolutely but it would be good if the addons world work (using the current api)

dfahlander:
Oops, I think I was too fast to merge it in. I see now that you have an unchecked box there. I will revert the merge.

dfahlander:
@chrahunt seems the PR couldn't be reopened again. Are you able to reopen it again?

dfahlander:
@chrahunt The reason I needed to revert the merge was that lots of tests started failing when applied. Something got really broken for some reason. Then I saw that there still was an unchecked box in your list so I assume you are still on merging from master and troubleshooting why tests fail. Maybe the PR could be reopened by you after pushing new commits to your branch. If not possible, please reopen a new PR.

dfahlander:
@chrahunt, would you be able to redo the PR?

chrahunt:
@dfahlander yup I see this now, thanks for the reminder.

Issue #409, opened 2016-12-21
RFC: Split Dexie.Synable into multiple modules


nponiros:
I think that it would be a good idea to split Dexie.Syncable into multiple smaller modules. With multiple modules it will be easier to write isolated unit tests for the various functions. I believe that it will be easier to get Dexie.Syncable out of alpha if we have isolated unit tests for its functions. I personally also find multiple smaller modules to be more readable and easier to understand than one large module.

From what I have seen, the build process can handle multiple modules so we shouldn't have an issue there.

Challenges:

* Many functions access variables which are global to the module or the parent function. This makes it harder to have those functions in their own module. We would have to pass all global variables as separate parameters or as options object
* I'm currently not sure how easy it will be to actually write isolated unit tests since Dexie.Syncable does not seem to use karma and the Dexie karma configuration only references the dexie file in the dist folder and not each individual file in the src directory. From what I can tell we would need to add a karma config which can reference the individual files

dfahlander:
You're 100% right. Should identify the private state vars and make them available somehow to splitted modules. Either via the db.syncable property, or some other strategy. Whatever seems most natural and easy to understand. Could use ES6 classes or even typescript if we would benefit from that.

Want to do the same for Dexie.js as well. At least cutting out Collection, Table and WhereClause for a start.

nponiros:
I don't think typescript would help us with Syncable but it can help with Dexie as we wouldn't need separate .d.ts files. Off course if Dexie is written in typescript then we should also write Syncable in typescript.

Cutting out Collection, Table and WhereClause and using classes for them sounds like a good ideas as it can fix https://github.com/dfahlander/Dexie.js/issues/341 and off course allow us to easier write isolated unit tests for those.

I could start working on Syncable and also add unit tests and we can later decide when to merge. I could open a long standing PR to better track the changes.

dfahlander:
That would be awesome. Dexie.Syncable / Dexie.Observable is a mess right now ;)

nponiros:
PR was merged. Closing this issue.

PR #413, opened 2016-12-21
WIP: Refactor Dexie.Syncable

nponiros:
This is an effort to refactor Dexie.Syncable and to split it into multiple smaller modules which can easier be unit tested.

There is still lots to do. Some of the things to do are:

- [x] Add an eslint configuration and eslint script
- [x] Try to find a better way to handle globals than the currently added globals module -> Use closures
- [x] Add karma configuration for travis (Dexie.Syncable)
- [x] Add shell script for travis to build Dexie/Dexie.Observable/Dexie.Syncable and run the tests/eslint
- [x] Add karma configuration for release (Dexie.Syncable)
- [x] Run Dexie.Syncable tests before release (karma release tests)
- [x] Add Dexie.Syncable integration tests to karma (at least the tests of tests-syncable.html)
- [x] Add html page for the Dexie.Syncable unit tests

Any feedback is appreciated. See also https://github.com/dfahlander/Dexie.js/issues/409


dfahlander:
👍 

Nice work. 

Maybe globals module not needed. Instead let those modules that needs to access or modify properties of db,  export a function that takes db as an argument.

nponiros:
I'm not sure what you mean exactly. Adjust the number of parameters of the exported functions or export functions receiving the globals and return a closure which will be the actual function executing the logic? I like the closure idea. We don't need to adjust the params of each function end we won't need a globals module.

dfahlander:
I was also thinking of the closure idea 👍

nponiros:
In order to get travis to run Dexie.Syncable tests on push we would have to add it to .travis.yml and also add the build call for Dexie.Syncable and probably also Dexie.Observable there. I'm not sure if the build tools will allow that in their current state (I don't know if the paths will match since we implicitly change the PWD).

The question for me is: Should we try to add Dexie.Syncable to .travis.yml or would it be better/easier to have Dexie.Syncable and possibly Dexie.Observable in their own repository? Those already have an own package.json and in some sense are treated as an own project.

dfahlander:
Good question. There's one benefit with keeping it in the same repo and require Travis tests for the addons to succeed - Changes to Dexie that would break the addons would be detected.

nponiros:
You are right about that. I don't know what will be more work. Possibly adjusting the build process or run tests (probably manually) on the syncable and observable repos each time Dexie is released or at least every major release. In the long run probably the latter. I like separate repositories in general but I get your point. In any case I wouldn't split the repos before Dexie v2.0.0 so I think I will try to add it to travis after I'm done refactoring.

dfahlander:
Sounds great. The release script (tools/release.sh) runs the unit tests on browserstack with a slightly larger set of target browsers than the one executed from Travis. It could also execute the addons tests. All releases including minor ones needs to pass those tests.

nponiros:
@dfahlander as you can see I made a great deal of changes and I think Dexie.Syncable is now more readable and also a bit more stable. I think this is enough refactoring for now. There are some more functions that could be put in an own module but those are very interconnected and I'm afraid that putting those in their own module will break more things than it will fix.

It would be great if you could review the changes when you get some time and after that (and after the tasks in the first comment are done) we could merge into master if you think that everything looks good.

dfahlander:
🥇 Impressive!

All tests runs perfectly and you've created lots of them. This was a huge refactoring. Simply amazing.

Seems all tasks are complete. Can I merge it in now?

nponiros:
Thanks. Regarding the merge: tests-syncable-dexie.html throws 13 exceptions in the browser but those are also there in the beta 7 tag so I don't know if those are relevant and what is missing is the addition of travis.sh to .travis.yml but maybe you can do that as you can probably better monitor the output of the script on travis. Other than that we are good to merge.

Issue #410, opened 2016-12-21
RFC: Make Dexie.Syncable MAX_CHANGES_PER_CHUNK configurable


nponiros:
Currently Dexie.Syncable uses 1000 as a threshold for when to send partial changes. This number is probably good enough in many cases but it is rather arbitrary and might cause issues if the object we have are large. It can also cause issues if the server used for the synchronization has a request size limit which is smaller than `1000 * object size`.

I think that it would be better if the user could define this number depending on the protocol used and the size of the objects in the database. I would suggest we extend the ISyncProtocol to allow a new property called `partialsThreshold` and let the user set this number to define when a partial change set should be sent. In case the user does not define the `partialsThreshold` property, we should use 1000 to ensure backward compatibility.

dfahlander:
True. It could also be an option provided in the `options` argument to [db.syncable.connect()](https://github.com/dfahlander/Dexie.js/wiki/db.syncable.connect()). Options can be read both by the framework and the sync protocol.

nponiros:
That could also be an option. The `options` probably also survive a reboot meaning that we don't have to configure the threshold every time. What worries me a bit with `options` is the fact the `connect()` could be called multiple times and from what I understand only the first call will actually set `options` the rest will use the existing connection. To change the options we would probably have to `disconnect()` and call `connect()` again.

We could probably even use the `context` to save the threshold and allow the user to change it any time he/she wants but that might become rather tricky.

dfahlander:
You may be right with your first suggestion to let the sync protocol offer an optional partialsThreshold. Maybe the default should be Infinity to simplify simple protocols.

Poll-based protocols that wants to optimize to a maximum could theoretically provide a really small treshold, measure what the JSON size would've been and return {again: 0} until it reaches a desired size. First when the server acks, provider could call ever onChangesAccepted() that corresponds to each chunk of changes.

PR #423, opened 2016-12-28
Make partials threshold configurable

nponiros:
Until now Dexie.Syncable sent partial changes when we had more than 1000
changes. With the partialsThreshold property this can be configured
during protocol registration.

Only positive numbers are allowed and if no threshold is given, Infinity
is used. We use Infinity so that simple server implementations of the
protocol don't have to care about partial changes.
If partialsThreshold is 0, Dexie.Syncable will send no changes but will
always set partial to true. This can be used to only get server changes
and ignore any client changes.

Closes https://github.com/dfahlander/Dexie.js/issues/410

### DIFF
--- a/addons/Dexie.Syncable/api.d.ts
+++ b/addons/Dexie.Syncable/api.d.ts
@@ -58,6 +58,7 @@ export enum DatabaseChangeType {
  * 
  */
 export interface ISyncProtocol {
+    partialsThreshold?: number;
     sync (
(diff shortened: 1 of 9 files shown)


dfahlander:
Perfect. I've updated the wiki at https://github.com/dfahlander/Dexie.js/wiki/Dexie.Syncable.ISyncProtocol

PR #424, opened 2016-12-28
Fixes for the ajax sync protocol

nponiros:
Fix some typos.

Add contentType property with application/json to the ajax call. Some
servers use it to know when to parse JSON.

Add warning that the needsResync flag is not implemented by
Dexie.Syncable.

Add partials flag to the server reply and pass it to
applyRemoteChanges().

When the server sends us a clientIdentity make sure we save it before
trying to resync. If saving failed we should stop syncing.


dfahlander:
👍

PR #428, opened 2016-12-29
WIP: Refactor observable

nponiros:
Long standing PR to refactor Dexie.Observable.

### DIFF
--- a/addons/Dexie.Observable/package.json
+++ b/addons/Dexie.Observable/package.json
@@ -29,6 +29,13 @@
   "scripts": {
     "build": "babel-node --presets es2015 --plugins syntax-async-functions,transform-regenerator tools/build.js",
     "watch": "babel-node --presets es2015 --plugins syntax-async-functions,transform-regenerator tools/watch.js",
+    "test": "npm run eslint && npm run build && npm run test:local",
+    "eslint": "eslint src",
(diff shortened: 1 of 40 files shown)


nponiros:
What is tests-observable-misc supposed to test? I'm guessing "test1" tests if the second database gets informed about changes in the first database. "Capital$$Id-test" seems to be testing if "$$Id" actually creates an ID.

There are also two functions in there (lines 85 and 170) which don't seem to be called ever. Should those be called? Unfortunately I can not tell what those are testing since most comments are not in english. Are those functions even relevant?

dfahlander:
Haven't touched those tests for ages. I think they might be very sloppy written. Sorry :(

nponiros:
Okay, I will see what I can do about it. Google translate might help :p

nponiros:
We should probably put sendMessage() and broadcastMessage() under the db.observable namespace since these are function defined by Dexie.Observable. This will also match what Dexie.Syncable does with the functions it defines.

nponiros:
So I think this is enough refactoring for now. I pulled some functions out of Dexie.Observable, added some unit tests and integration tests and I also added the tests to travis.sh and release.sh.

There is still more that we could refactor but I will not have so much time in the next few months and I don't want to start something which I can not finish. If everything looks good feel free to merge @dfahlander .

Issue #447, opened 2017-01-15
RFC: Change a syncNode's options without having to delete it


nponiros:
From what I can tell the only way to change the options of a remote syncNode is to delete the node and recreate it with connect(). During connect if a node with the given URL is found, the options are just ignored. I don't know if there is a reason for this behavior. Deleting the node seems to be problematic as that could delete all the changes of the node that were not synchronized yet and it does delete all changes in the \_uncommittedChanges table.

One possible solution for this would be to not ignore the options if `connect()` is called on an already existing node. From what I can tell the node would have to be disconnected and reconnected for the new options to take effect and be passed to the `sync()` function. This should be documented in the wiki if we choose this solution.

Another solution would be to have a new method `setOptions(url, options)` which would automatically disconnect and reconnect the node for the options to take effect.

Furthermore it would be nice to have a method `getOptions(url)` which would return the options for a syncNode. This is not so important as there are other ways to get the options if needed but I think it would be cleaner/safer if Syncable offers this method, especially if we have a `setOptions()` method.

The reason why I would like to be able to set and get the options is because I want to allow a user of my app to configure the options without having to store the options myself or having to delete the node to be able to set new options.

dfahlander:
connect() should not ignore the options I think, no matter if we'd also support setOptions() and getOptions(). You're right that a connect() with different options should reconnect the node.

dfahlander:
I'm ok with adding setOptions() and getOptions(). But I think that connect() with alternate options not reconnecting is a bug and should also be solved.

nponiros:
Okay, thanks for the input @dfahlander . Will try to fix/implement this asap.

PR #459, opened 2017-01-28
Change options for an existing node

nponiros:
This commit makes two changes.

1. The connect() method was changed to be able to set options for an
existing node. If the node is disconnected then the options will just be
saved on the existing node. If the node is already connected, it will be
disconnected, the options will be saved and the node will be
reconnected. This only happens if the current options for the node
differ from the options connect() was called with.

2. A new method getOptions(url) was added which can be used to get the
current options for the given node.

This commit closes https://github.com/dfahlander/Dexie.js/issues/447

I decided not to implement a setOptions() method as I don't think that
it has much usage if we can have the same effect by using connect().


nponiros:
I managed to add the missing tests to the unit tests. Unfortunately I had to copy the sample websocket protocol into the test in order for it to work. I tried using it from the sample directly as it was before but the test would hang because the Syncable instance in the test did not have a registered protocol. No clue why this was the case but copying the sample protocol into the test fixed it.

I also noticed some differences in the package.json of the addons/Dexie. For example the Dexie.Syncable package.json does not have a module entry. Also the build processes are differently setup (but probably do the same in the end with different syntax). Also some folder names are different and it seems that some of the config files are not used. I could try to eliminate this changes in a different PR.

dfahlander:
There's a [wiki page](https://github.com/dfahlander/Dexie.js/wiki/Dexie.Syncable.js) that mirrors the README with getOptions() that mirrors the README. Updated that one as well.

nponiros:
Thanks, missed that one ;)

PR #518, opened 2017-04-28
Update typescript Promise definition

SteeluserTeam:
Update Promise from this file https://github.com/Microsoft/TypeScript/blob/v2.1.4/src/lib/es2015.promise.d.ts because the current definition seems not compatible with typescript 2.3 PromiseLike definition.
For example :
```typescript
await Promise.all([dexiePromise1, dexiePromise2])
```
does not currently return the same type between typescript 2.2.2 and 2.3.1 without this change.


dfahlander:
I earlier today went on the same issue in branch [upgrade-typescript-to-2-3-1](https://github.com/dfahlander/Dexie.js/tree/upgrade-typescript-to-2-3-1). But your PR is better as it stays backward compatible.
Thanks! 👍

Issue #567, opened 2017-08-02
How to use 'dexie-observable' and 'dexie-relationships' addons?


5uper:
I am using Dexie.js in my ionic2 project and I am trying to add **'dexie-observable'** and **'dexie-relationships'** together **in one Database**, like so:

```
import Dexie from 'dexie';
import 'dexie-observable';
import relationships from 'dexie-relationships';
...

export class DataBase extends Dexie {

    public matches: Dexie.Table<MatchModel, number>;

    constructor() {
      super("UserDB", { addons: [relationships] });
...
```

the problem is  'dexie-relationships' **overrides** the 'dexie-observable' addon, **is it possible to use both addons in the same database?**

i tried also this:

```
import Dexie from 'dexie';
import dexieObservable from 'dexie-observable';
import relationships from 'dexie-relationships';
...

export class DataBase extends Dexie {

    public matches: Dexie.Table<MatchModel, number>;

    constructor() {
      super("UserDB", { addons: [relationships, dexieObservable] });
...
```

or

```
...
import Dexie from 'dexie';
import dexieObservable from 'dexie-observable';
import relationships from 'dexie-relationships';

Dexie.addons.push(dexieObservable);
...
```

used versions:
```
"dexie": "2.0.0-beta.11"
"dexie-observable": "1.0.0-beta.3"
"dexie-relationships": "1.2.8"
```

5uper:
solved with [#568](https://github.com/dfahlander/Dexie.js/pull/568) pull request

PR #568, opened 2017-08-02
fix typescript (type error) on adding addons

5uper:
solves problem in [#567](https://github.com/dfahlander/Dexie.js/issues/567)

Typescript expected typeof `(db: Dexie) => void`

```
import Dexie from 'dexie';
import dexieObservable from 'dexie-observable';
import relationships from 'dexie-relationships';
...

export class DataBase extends Dexie {

    constructor() {
      super("YourDB", { addons: [relationships, dexieObservable] });
...
```

### DIFF
--- a/addons/Dexie.Observable/src/Dexie.Observable.d.ts
+++ b/addons/Dexie.Observable/src/Dexie.Observable.d.ts
@@ -113,4 +113,6 @@ declare module 'dexie' {
     }
 }
 
-export default Dexie.Observable;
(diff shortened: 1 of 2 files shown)


dfahlander:
I'm afraid the changes would make methods on dexieSyncable disappear, like dexieSyncable.registerSyncProtocol(). Or am I wrong?

dfahlander:
Thanks for the PR. I had to adjust it to allow the static APIs of the addons to be available as well. Replaced by #580

Issue #569, opened 2017-08-03
Dexie.Syncable getting stuck when given a large number of changes


e45lee:
Dexie.Syncable gets stuck when trying to apply a large number of changes.  I've stepped through it in the debugger and it specifically gets stuck here (lines 28-30 in https://github.com/dfahlander/Dexie.js/blob/master/addons/Dexie.Syncable/src/finally-commit-all-changes.js):
```
      }).then(function () {
        // 3. Apply last chunk of changes
        return applyChanges(changes, 0);
      }
```
The database appears to apply all the changes (and applyChanges is called the appropriate number of times -- verified by console.log) but we never get to the next step (lines 31-33)
```
      }).then(function () {
        // Get what revision we are at now:
        return db._changes.orderBy('rev').last();
      }).then(function (lastChange) {
```

e45lee:
FYI, we see this issue on Chrome 59 (will try 60 when I have a chance).  The same change set, when applied in Firefox, works fine.

e45lee:
It seems that there's some strange Dexie.Promise behaviour when adding a whole bunch of items to IndexedDB via chained promises -- and applyChanges in syncable is implemented via chained promises.  The following bit of code never gets to the transaction completed stage nor the friends added stage,
and no error is thrown.

```
<html>
  <body>
    <script src="https://unpkg.com/dexie@^2.0.0-beta/dist/dexie.js"></script>
    <script>
    var db = new Dexie("FriendDatabase");
    db.version(1).stores({
      friends: "++id,name,age"
    });

    function addFriend(n) {
      if ( n % 1000 === 0 )
        console.log(`Adding friend ${n}`);
      if (n === 0)
        return true;
      else {
        return db.friends.bulkAdd([{name: `Foo${n}`, age: n}])
                .then(() => addFriend(n-1));
      }
    }

    function iteratedAddFriend(n) {
      while(--n) {
        db.friends.bulkAdd([{name: `Foo${n}`, age: n}]);
        if ( n % 1000 === 0 )
          console.log(`Adding friend ${n}`);
      }
      return Dexie.Promise.resolve(true);
    }

    db.transaction('rw', db.friends, () => {
      return addFriend(20000).then(() => {
        console.log("Friends added!");
      }).catch((e) => {
        console.log(`Friend error: ${e}`);
      });
    }).then(() => {
      console.log("Transaction completed!");
    }).catch((e) => {
      console.log(`Friend error: ${e}`);
    });

    </script>
[...]

PR #570, opened 2017-08-05
Change apply-changes in Dexie.Syncable to use a transaction to manage changes instead of chained promises

e45lee:
This pull request changes apply-changes to use a transaction to manage changes instead of using
chained promises.

This is to mitigate #569 as promises seem to behave badly in the presence of a large number (~18k) of changes.

### DIFF
--- a/addons/Dexie.Syncable/src/apply-changes.js
+++ b/addons/Dexie.Syncable/src/apply-changes.js
@@ -3,15 +3,11 @@ import { CREATE, DELETE, UPDATE } from './change_types';
 import bulkUpdate from './bulk-update';
 
 export default function initApplyChanges(db) {
-  return function applyChanges(changes, offset) {
-    if (offset === changes.length)
-      return Dexie.Promise.resolve(null);
-    changes = changes.slice(offset);
-
+  return function applyChanges(changes) {
     let collectedChanges = {};
-    changes.map((change) => {
+    changes.forEach((change) => {
       if (!collectedChanges.hasOwnProperty(change.table)) {
-        collectedChanges[change.table] = [[], [], [], []];
+        collectedChanges[change.table] = { [CREATE]: [], [DELETE]: [], [UPDATE]: [] };
       }
       collectedChanges[change.table][change.type].push(change);
     });
@@ -19,7 +15,7 @@ export default function initApplyChanges(db) {
     let tables = table_names.map((table) => db.table(table));
 
     return db.transaction("rw", tables, () => {
-      table_names.map((table_name) => {
+      table_names.forEach((table_name) => {
         const table = db.table(table_name);
         const specifyKeys = !table.schema.primKey.keyPath;
         const createChangesToApply = collectedChanges[table_name][CREATE];
--- a/addons/Dexie.Syncable/src/apply-changes.js
+++ b/addons/Dexie.Syncable/src/apply-changes.js
@@ -4,31 +4,35 @@ import bulkUpdate from './bulk-update';
 
 export default function initApplyChanges(db) {
   return function applyChanges(changes, offset) {
-    const length = changes.length;
-    // This is the base case for the recursion
-    if (offset >= length) return Dexie.Promise.resolve(null);
-    const firstChange = changes[offset];
-    let i, change;
-    for (i=offset + 1; i < length; ++i) {
-      change = changes[i];
-      if (change.type !== firstChange.type ||
-          change.table !== firstChange.table)
-        break;
-    }
-    const table = db.table(firstChange.table);
-    const specifyKeys = !table.schema.primKey.keyPath;
-    const changesToApply = changes.slice(offset, i);
-    const changeType = firstChange.type;
-    const bulkPromise =
-        changeType === CREATE ?
-            table.bulkPut(changesToApply.map(c => c.obj), specifyKeys ?
-                changesToApply.map(c => c.key) : undefined) :
-            changeType === UPDATE ?
-                bulkUpdate(table, changesToApply) :
-                changeType === DELETE ?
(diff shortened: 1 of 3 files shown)


addons/Dexie.Syncable/src/apply-changes.js
```diff
@@ -4,31 +4,35 @@ import bulkUpdate from './bulk-update';
 
 export default function initApplyChanges(db) {
   return function applyChanges(changes, offset) {
```
nponiros: I don't think offset is needed anymore. applyChanges is only called with 0 from what I can tell. Offset was needed for the recursion which we no longer have.

addons/Dexie.Syncable/src/apply-changes.js
```diff
@@ -4,31 +4,35 @@ import bulkUpdate from './bulk-update';
...
-                    Dexie.Promise.resolve(null);
+    if (offset === changes.length)
+      return Dexie.Promise.resolve(null);
+    changes = changes.slice(offset);
 
-    return bulkPromise.then(()=>applyChanges(changes, i));
+    let collectedChanges = {};
+    changes.map((change) => {
+      if (!collectedChanges.hasOwnProperty(change.table)) {
+        collectedChanges[change.table] = [[], [], [], []];
```
nponiros: Can you please use an object here instead of an array? I don't see how an array is better than an object here. Maybe something like this:
```js
collectedChanges[change.table] = { [CREATE]: [], [DELETE]: [], [UPDATE]: []};
```

addons/Dexie.Syncable/src/apply-changes.js
```diff
@@ -4,31 +4,35 @@ import bulkUpdate from './bulk-update';
...
-                changeType === DELETE ?
-                    table.bulkDelete(changesToApply.map(c => c.key)) :
-                    Dexie.Promise.resolve(null);
+    if (offset === changes.length)
+      return Dexie.Promise.resolve(null);
+    changes = changes.slice(offset);
 
-    return bulkPromise.then(()=>applyChanges(changes, i));
+    let collectedChanges = {};
+    changes.map((change) => {
```
nponiros: Using forEach would be better here in my opinion as you don't care for the return value of  map.

addons/Dexie.Syncable/src/apply-changes.js
```diff
@@ -4,31 +4,35 @@ import bulkUpdate from './bulk-update';
...
+      if (!collectedChanges.hasOwnProperty(change.table)) {
+        collectedChanges[change.table] = [[], [], [], []];
+      }
+      collectedChanges[change.table][change.type].push(change);
+    });
+    let table_names = Object.keys(collectedChanges);
+    let tables = table_names.map((table) => db.table(table));
+
+    return db.transaction("rw", tables, () => {
+      table_names.map((table_name) => {
```
nponiros: Using forEach would be better here in my opinion as you don't care for the return value of  map.

addons/Dexie.Syncable/src/apply-changes.js
```diff
@@ -4,31 +4,35 @@ import bulkUpdate from './bulk-update';
...
+    changes.map((change) => {
+      if (!collectedChanges.hasOwnProperty(change.table)) {
+        collectedChanges[change.table] = [[], [], [], []];
+      }
+      collectedChanges[change.table][change.type].push(change);
+    });
+    let table_names = Object.keys(collectedChanges);
+    let tables = table_names.map((table) => db.table(table));
+
+    return db.transaction("rw", tables, () => {
```
nponiros: I'm not sure if we need a transaction here as we should already be in a transaction scope from finally-commit-all-changes but @dfahlander should be the judge of that.

e45lee: I put a transaction here as it looked like finally-commit-all-changes depended on either all of the operations happen or they don't.

addons/Dexie.Syncable/src/apply-changes.js
e45lee: No problem.

addons/Dexie.Syncable/src/apply-changes.js
e45lee: No problem.

addons/Dexie.Syncable/src/apply-changes.js
e45lee: No problem.

addons/Dexie.Syncable/src/apply-changes.js
e45lee: No problem.

e45lee:
@nponiros I've made the changes you've requested.

nponiros:
@e45lee Thanks for the fixes. Lets see what @dfahlander says. If he is happy with the PR, he will merge it.

e45lee:
@nponiros Do you know when @dfahlander can take a look at this?  There's no rush, it's just that I need to know if I have to host our fixes temporarily locally for the time being -- we're deploying software in Fall '17 that relies on Syncable, and this would be a problem that I can see our students running into.

Thanks again.

nponiros:
@e45lee unfortunately no.. He hasn't been very active the last few months. I guess hosting a fixed version for your project is a good idea.

e45lee:
@dfahlander Will you have a chance to look at this PR soon?

dfahlander:
Yes! Sorry for not having the time until now. I hope to get some time today. Otherwise next week at latest.

dfahlander:
I'm testing out this PR on my own application. Will review changes as well tomorrow and merge if all goes fine.

dfahlander:
Seems to be working well. Thanks!

Issue #565, opened 2017-07-23
SecurityError on Safari


gigamorph:
Hello,

I have an application that uses Dexie and it is failing on Safari (the desktop version). I'm pretty sure it worked before but recently found out about it by examining user complaints. 

The fiddle from the Dexie homepage https://jsfiddle.net/dfahlander/3tf5r0cu/ also produces error in the Log: 
```
Using Dexie v1.5.1
SeurityError - This browser doesn't like fiddling with indexedDB.
Go run some samples instead at: https://github.com/dfahlander/Dexie.js/wiki/Samples
```

I also put the example code from the GitHub README (which is mostly the same code) to my CodePen: https://codepen.io/gmorph/pen/gxObEz?editors=1001 and you can (hopefully) see the alert popup says 

```
Error: OpenFailedError: SecurityError IDBFactory.open() called in an invalid security context
```

It works fine on Chrome, FireFox, MS Edge (in VirtualBox), and Opera instances that I have on my MacBook Pro, but only Safari (Version 10.1.2 (11603.3.8)) is failing on me.

Hopefully it will ring a bell with the core developers :) Thank you.

nponiros:
Hey there, I don't know this error but I can try to reproduce it tomorrow. Have you tested with dexie v2 also? Could it be an issue with codepen and jsfiddle? Maybe those are running the code in an iframe which causes issues?

gigamorph:
My apologies. While creating the CodePen example I totally forgot about the original problem. Yes, I think there's a (probably known) problem of running the code with codepen and jsfiddle, I would also appreciate explanation on that problem, but that wasn't really what made me come here. After some fiddling the problem seems to be that I was trying to save an object that is not serializable (with a link to a class instance with functions and possibly some recursive data), and that might have caused the error. Which now leaves me wondering, how does it work on other browsers then?

nponiros:
Could it be that other browsers failed silently? It might also be a browser bug, not the first time browsers implement things differently.

dfahlander:
I am not 100% sure, but I've always supposed the problem has been that Safari forbids to access IDB from an iframe.

sechel:
Thats right, it such a problem that we need to open a separate tab just to support safari. Extremely ugly.

lsentkiewicz:
I also have a problem with Safari.
I have a webworker with following code (inside a class):
```js
    this.db = new Dexie('db');
    this.db.version(1).stores({
      events: 'serialNumber',
    });
[...]

PR #579, opened 2017-08-28
Workaround for Safari issue with getAll().

dfahlander:
Resolves issue #565

### DIFF
--- a/src/Dexie.js
+++ b/src/Dexie.js
@@ -2785,7 +2785,8 @@ export default function Dexie(dbName, options) {
         for (var i = 0; i < storeNames.length; ++i) {
             var storeName = storeNames[i];
             var store = idbtrans.objectStore(storeName);
-            hasGetAll = 'getAll' in store;
+            hasGetAll = navigator.userAgent.indexOf("Safari") === -1 && // See issue #565
+                'getAll' in store;
             for (var j = 0; j < store.indexNames.length; ++j) {
                 var indexName = store.indexNames[j];
                 var keyPath = store.index(indexName).keyPath;


timbru31:
This breaks the performance of e.g. Chrome, too and even the Safari (iOS 10.3, iOS 11) versions which do support IndexdedDB 2.0 and the *getAll*  method.
Chrome user agent: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36`

See my [stackoverflow question/issue ](https://stackoverflow.com/q/46475404/1902598) regarding very, very bad performance. Testing with getAll instead results in 10x faster loading times for us.
Please consider reverting this issue.

dfahlander:
@timbru31 Thanks. I never like UA sniffing, but need to find a way to work around the bug of Safari idintified in [the more recent comments of issue 565](https://github.com/dfahlander/Dexie.js/issues/565#issuecomment-324679947). If I roll back this PR we will get issues again on Safari: Version 10.1.2 (12603.3.8).

Any idea someone?

nponiros:
Hm maybe `navigator.vendor` can help? On Mac OS Safari the vendor is "Apple Computer, Inc.". You should be able to detect Safari with this but you will probably have issues on iOS Chrome which will probably (didn't test) give vendor "Google Inc." although parts of Safari are used under the hood.

timbru31:
Besides that, getAll works fine for me on iOS 10.3 and iOS 11 - I'd like to keep support for these, too.

nponiros:
We might be able to combine the vendor with the version from the userAgent but the test will probably be rather brittle..

timbru31:
navigator.appVersion might help to detect the iOS versions: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/appVersion (deprecated though :/)

nponiros:
appVersion on Mac OS Safari is basically the same as userAgent minus the "Mozilla" at the beginning of the UA string..

dfahlander:
The issue in Safari 10.1 on iOS 10.3 is rather severe. Repro: http://dexie.org/test/dexie-issue-565/

dfahlander:
Output: Testing Safari issue with IDBObjectStore.getAll()

Status: Preventing eternal reload. Last status was: Could put item. Now trying IDBObjectStore.getAll()

navigator.userAgent: Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X) AppleWebKit/603.1.30 (KHTML, like Gecko) Version/10.0 Mobile/14E269 Safari/602.1

navigator.vendor: Apple Computer, Inc.

navigator.appVersion: 5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X) AppleWebKit/603.1.30 (KHTML, like Gecko) Version/10.0 Mobile/14E269 Safari/602.1

dfahlander:
Also on desktop mac, same issue:

navigator.userAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/603.3.8 (KHTML, like Gecko) Version/10.1.2 Safari/603.3.8

navigator.vendor: Apple Computer, Inc.

navigator.appVersion: 5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/603.3.8 (KHTML, like Gecko) Version/10.1.2 Safari/603.3.8

dfahlander:
Could someone try [the repro](http://dexie.org/test/dexie-issue-565/) on a newer Safari version to see if it is solved? Note that this happens from a WebWorker calling IDBObjectStore.getAll().

Unless we can see a non-failing Safari version, we need to keep preventing using getAll() on Safari. If getting (but keep Chrome untouched of course)

timbru31:
iPad with iOS 11.1 (15B5066f) is fine:
img_b81283582225-1

---
iPhone with iOS 11.0.1 (15A402) is fine

img_3395740420d7-1

timbru31:
However, I noticed that running the code outside a Worker, the web page did not reload - this means we need to detect a Worker environment and the special Safari versions?

dfahlander:
Ok, great, so Safari 604 is fine but not <603. What about the following:

```js
hasGetAll = 'getAll' in store &&
                !(/Apple/.test(navigator.vendor) &&
                  [].concat(navigator.userAgent.match(/Safari\/(\d*)/))[1] < 604);
```

timbru31:
What about the WebWorker check, too? Something like this could work: https://stackoverflow.com/a/18002694/1902598

dfahlander:
I haven't checked that it's only an issue in workers. I'll test now.

dfahlander:
You're right. I can fix it only for workers. But I found that navigator.vendor is not accessable from workers :(  only navigator.userAgent seem to be accessable. So I have to do something else than /Apple/.test(navigator.vendor). Suggestion?

dfahlander:
I've updated the check to:

```js
// Bug with getAll() on Safari ver<604 on Workers only, see discussion following PR #579
if (/Safari/.test(navigator.userAgent) &&
    !/Chrome/.test(navigator.userAgent) &&
    _global.WorkerGlobalScope && _global instanceof _global.WorkerGlobalScope &&
    [].concat(navigator.userAgent.match(/Safari\/(\d*)/))[1] < 604)
{
    hasGetAll = false;
}    
```
Thoughts?

dfahlander:
Ah! Edge also has "Safari" in UA! Another false positive. Need to update yet again.

```js
if (/Safari/.test(navigator.userAgent) &&
    !/(Chrome\/|Edge\/)/.test(navigator.userAgent) &&
    _global.WorkerGlobalScope && _global instanceof _global.WorkerGlobalScope &&
    [].concat(navigator.userAgent.match(/Safari\/(\d*)/))[1] < 604)
{
    hasGetAll = false;
}    
```

Sopamo:
Sorry for opening this issue up again. I'm pretty sure I'm running into this issue again on an iPhone SE iOS 10.3 emulator using a webkit view (cordova). `navigator.userAgent` reports: 
`Mozilla/5.0 (iPhone; CPU iPhone OS 10_3_1 like Mac OS X) AppleWebKit/603.1.30 (KHTML, like Gecko) Mobile/14E8301` which doesn't contain "Safari", so the user agent sniffing doesn't work in this case. I'm currently unable to check if the issue only affects emulated devices because I do not own an iOS device running iOS 10. I'd be pretty surprised if the user agent differs between a real device and an emulator though.
Upon calling table.toArray() the app just crashes. Loading a single item via .get() works fine.

dfahlander:
If you use dexie@3 it will work the same on all browsers.

Issue #589, opened 2017-09-30
Maximum call stack exceeded when running Dexie with zone.js 0.8.18


duydao:
I'm getting the following error when accessing any database using Ionic 3.7:

```
dexie.es.js:1382 Unhandled rejection: RangeError: Maximum call stack size exceeded
    at getErrorWithStack (http://localhost:8100/build/vendor.js:120381:27)
    at new Promise (http://localhost:8100/build/vendor.js:120807:29)
    at Promise.then [as __zone_symbol__then] (http://localhost:8100/build/vendor.js:120836:22)
    at r (http://localhost:8100/build/polyfills.js:3:7143)
    at t.scheduleTask (http://localhost:8100/build/polyfills.js:3:14860)
    at r.scheduleTask (http://localhost:8100/build/polyfills.js:3:10785)
    at r.scheduleMicroTask (http://localhost:8100/build/polyfills.js:3:11037)
    at f (http://localhost:8100/build/polyfills.js:3:19578)
    at t.then (http://localhost:8100/build/polyfills.js:3:21934)
    at http://localhost:8100/build/vendor.js:120730:41
```

The message keeps repeating in the console.

I've tested the following things:

- The Error does not occur when I'm using the "Getting started" example without typescript
- It works with ionic 3.6 using Angular 4.5.3 and zone.js 0.8.18

I've reproduced the problem by creating a fork from the ionic conference example app, which be used like this:

```
git clone https://github.com/duydao/ionic-conference-app.git
npm i
npm run serve
```
- open `localhost:8100` with chrome when the server is ready
- call `window.dexieService.friends.count().then(console.log).catch(console.error)` in the dev console
- error occures

The dexie implementation can be found in `src/pages/schedule/schedule.ts` and is as simple as this:

```
export class DexieService extends Dexie {
  friends: Dexie.Table<any, string>;

  constructor() {
    super('friend_database');
    this.version(1).stores({
      friends: 'name,shoeSize'
    });
  }
}
```

any hints on how I can debug this error?

duydao:
- This issue may be related to https://github.com/angular/zone.js/issues/886
- Running Ionic 3.7 with zone.js 0.8.18 works as well. Seems to be a problem with ionic 3.7.1 and zone.js 0.8.18

duydao:
[...]

PR #590, opened 2017-10-01
Fixing incompability with zonejs 0.8.18

dfahlander:
Trying to fix issue with zonejs 0.8.18 #589


duydao:
This patch works for me with angular 4.4.4 and zone.js 0.8.18 in Chrome 61, Safari 10 and even in FF 57.0b3 on Mac, thanks!

PR #591, opened 2017-10-02
Partially revert pr 579

dfahlander:
PR #579 was meant to work around a severe Safari issue with IDBObjectStore.getAll() that when called from a Web worker, results in a page crash followed by an automatic page reload. Since getAll() is just an optimization and not needed for IndexedDB to work, PR 579 shut down this optimization of utilizing on Safari. Due to the nature of the Safari bug (page crash + reload), feature testing was not possible - we could only do user-agent sniffing to avoid the bug.

The user-agent sniffing in PR  #579 did only look for "Safari" in user-agent, which made it also affect Chrome and Opera browsers (which was not the intent). Also, the bug has since been fixed in a later version of Safari + it only happened when called from a web worker, so we needed a more fine grained control of whether the getAll() optimization should be avoided or not. 

This new PR will replace the user-agent code from PR #579 with another more fine grained check that only disables getAll() when:

* browser is Safari (user agent contains "Safari" but not "Chrome" or "Edge")
* we are in a Worker environment
* Safari internal version number is less than 604, as the issue has been fixed in 604.

### DIFF
--- a/src/Dexie.js
+++ b/src/Dexie.js
@@ -2787,8 +2787,9 @@ export default function Dexie(dbName, options) {
         for (var i = 0; i < storeNames.length; ++i) {
             var storeName = storeNames[i];
             var store = idbtrans.objectStore(storeName);
-            hasGetAll = navigator.userAgent.indexOf("Safari") === -1 && // See issue #565
-                'getAll' in store;
+            hasGetAll = 'getAll' in store &&
+                !(/Apple/.test(navigator.vendor) &&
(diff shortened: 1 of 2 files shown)


timbru31:
Thanks for the quick fix and debugging help. While UA sniffing isn't a great thing, this looks like a good fix for the issue - hopefully Safari 10.1 / iOS 10.3 sees another patch release which takes care of this WebKit bug.

PR #628, opened 2017-12-27
Update README.md

everdimension:
Since v2 is no longer in beta, unnecessary beta mentions should be removed


dfahlander:
Thanks 👍

PR #671, opened 2018-03-03
fix for removing array items

bozzaj:
Using Dexie.delByKeyPath(), which ultimately calls Dexie.setByKeyPath(), causes some problems when the path ends in an array.

Removing an item from an array using delete does not updating the array length. Instead it creates an undefined array item, leaving the length alone.

This fix identifies the object as an Array and uses Array.splice() to remove the keyPath.

See here: [delete operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete) under Deleting Array Elements.


bozzaj:
closing this pull request to create another one properly

PR #672, opened 2018-03-03
fix for setByKeyPath when removing array items

bozzaj:
I ran into an issue (using 2.0.2) with Dexie.Observable. When receiving the UPDATED changes event, the object returns was not the same as the object in the database. Tracing it down, I found an issue with the way deletes were applied.

Using Dexie.delByKeyPath(), which ultimately calls Dexie.setByKeyPath(), causes some problems when the path ends in an array.

Removing an item from an array using delete does not updating the array length. Instead it creates an undefined array item, leaving the length alone. Essentially, this means that:

```
const obj = {deepObject: {someArray: ["a", "b", "c"]}};
Dexie.delByKeyPath(obj, "deepObject.someArray.2");
```
results in:
```
obj === {deepObject: {someArray: ["a", "b", null]}};
```
instead of 
```
obj === {deepObject: {someArray: ["a", "b"]}};
```

This fix identifies the object as an Array and uses Array.splice() to remove the keyPath.

See here: [delete operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete) under Deleting Array Elements for additional understanding of the problem.

### DIFF
--- a/src/functions/utils.ts
+++ b/src/functions/utils.ts
@@ -141,14 +141,20 @@ export function setByKeyPath(obj, keyPath, value) {
             var currentKeyPath = keyPath.substr(0, period);
             var remainingKeyPath = keyPath.substr(period + 1);
             if (remainingKeyPath === "")
-                if (value === undefined) delete obj[currentKeyPath]; else obj[currentKeyPath] = value;
-            else {
+            if (value === undefined) {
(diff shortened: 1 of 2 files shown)


bozzaj:
small additional fix for wrong indent.

PR #675, opened 2018-03-08
Updated dts-bundle-generator

timocov:
Hi there!

[I just released](https://github.com/timocov/dts-bundle-generator/releases/tag/v1.0.0) new version of `dts-bundle-generator` where I have fixed the bug when `export default` is removed from output - so you do not need to append it by yourself anymore 🙂 .

As soon `package-lock.json` is OS-dependent I am not sure how to update it right way. If you want - I can just push all changes made by my local `npm`.


dfahlander:
Thanks a lot 👍

PR #681, opened 2018-03-13
Make sure navigator is available in our context before using it

remusao:
Hi,

I recently encountered an exception where Dexie would try to access `navigator`, which was not available. I see that in the codebase it is usually checked that `navigator` is available before using, except in one place. So I added this check and it solves my issue.

I did not understand 100% the issue, but it seemed to happen while initializing a new version of a Dexie database with an extra table.

Cheers,
Rémi


dfahlander:
Thanks a lot. Important change as Dexie can be run on node with IndexedDBShim.

PR #686, opened 2018-03-30
Update the readme.md to support 2.7 with strictPropertyInitialization

MrDesjardins:
TypeScript 2.7 with strictPropertyInitialization to true require specifying that a member will be initialized later (lazy-initialized). This is the case and it won't compile without the exclamation point to anyone having this strict policy turned on. The code will also work for people with the flag off. It would be clearer if the readme had it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,9 +157,9 @@ interface IFriend {
 // Declare Database
 //
 class FriendDatabase extends Dexie {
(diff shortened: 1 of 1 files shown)


LewisAndrewCampbell:
As a new user this confused me as well, TIL about "!" on typescript properties.

dfahlander:
Manually merged this PR.

Issue #701, opened 2018-05-25
Get tests running under Node.js with IndexedDBShim


aral:
It’s possible to run Dexie under Node using IndexedDBShim (https://github.com/dfahlander/Dexie.js/issues/359#issuecomment-390636035), however, we do not know for certain that it behaves as expected as the test suite currently only runs in the browser environment.

Given that IndexedDBShim currently has [some issues under Node](https://github.com/axemclion/IndexedDBShim/issues/296), it would be good to now whether and how it effects expected behaviour in Dexie.

@dfahlander Do you have any thoughts on a strategy to make the test suite isomorphic so that it can be run under both browsers and Node?

aral:
Just a quick update that I’m working on this. Just a brief update:

Made some progress using [node-qunit](https://github.com/qunitjs/node-qunit) but that wasn’t picking up the async tests (it’s not actively maintained).

I just got the first batch of test (in tests-table.js) to run under Node using qunit-tap (after seeing [this comment](https://github.com/qunitjs/node-qunit/issues/110#issuecomment-172603925) by @leobalter – thanks, Leo.)

Currently, it’s a terrible hack (I was just trying to get it working) and I’ve hardcoded some browser-based globals that _dexie-unittest-utils.js_ was setting, etc. I also replaced the imports with requires to get it working.

Next step: going to try and make a test runner that doesn’t require any code changes on the existing tests. Will keep you updated on my progress and push a brach to my fork when I have something a bit more presentable.

PS. Currently, all tests in tests_table.js are passing under Node, except for test 5 (`not ok 5 - Transaction failed: ReferenceError: IDBKeyRange is not defined, expected: true, got: false, test: where, module: table`) and 17 (`not ok 17 - ReferenceError: IDBKeyRange is not defined, expected: true, got: false, test: put, module: table`). Will take a quick look into those also.

dfahlander:
Thanks a lot for looking into this. The tests will definitely need to be converted to universal JS in one or other way. I just looked up the issue with IDBKeyRange, and found a few locations in WhereClause. Should be the only ones using it. It's now fixed in master.

aral:
Awesome news on the IDBKeyRange issues. Going to start working on a very minimalist test runner for Node now: basically, my plan is to:

1. Update the tests themselves so any browser-specific  globals are conditionally applied or replaced with defaults that make sense for Node.
2. Concatenate all the tests into a single file.
[...]

PR #707, opened 2018-05-28
Node tests

aral:
This is a first attempt at getting Dexie tests running (and eventually passing) under Node.js.

To use:

## Raw [TAP](http://testanything.org/tap-specification.html) output

(Verbose; more informative, harder to skim.)

```bash
npm run test:node
```

## Pretty output

(Summary view courtesy of [faucet](https://github.com/substack/faucet).)

```bash
npm run test:node:pretty
```

## Details

I’ve added a very simple custom test runner that does the minimum it needs to set up an environment in which existing tests can run. I’ve updated the test utilities to make them isomorphic (was just a couple of variables in a couple of places).

Specifically, it:

  * Creates a CommonJS version of the dexie-unittest-utils script
  * Sets up a Node testing environment using IndexedDBShim, QUnit, and QUnit-tap
  * Loads test scripts, strips imports from them and runs them in the Node environment

Each script is run in its own process so that failures do not affect the others. (At this point, I thought it would be more important to give us an initial overview of the state of the tests under Node.js instead of discovering it piecemeal. We can review this decision in the future.)

**Current status:** the runner executes all of the tests apart from those within specific test scripts following a timeout. It does, however, make its way through all the test scripts. Some test and some test scripts are timing out. Individual test timeouts are currently set to 3 seconds and test script timeouts at 10 seconds. Some of the issues may be due to issues with the test environment I’m setting up.

  * @dfahlander’s initial analysis and summary: https://github.com/dfahlander/Dexie.js/issues/701#issuecomment-392756378
  * Full list of failures based on these tests: https://github.com/dfahlander/Dexie.js/issues/709


aral:
@dfahlander I just reformatted the code to match the project style used in the other tests. Barring any changes you’d like me to make, I believe this is now ready to be merged.

dfahlander:
Closing this for cleanup purpose.

PR #731, opened 2018-07-20
Allow accessing to-be-deleted stores in upgraders (#105)

sim642:
My attempt at making to-be-deleted stores still available via upgrader's transaction by specifically putting them back into a special upgrader schema. Not sure if this is the right approach but the added tests for shorter table rename and primary key change seem to pass with it.


dfahlander:
Thanks. Looking at the code, this seems correct. And as all tests pass and you provided great unit tests, seems fine to merge in.

Thanks a lot for taking the action and do the PR!

Issue #606, opened 2017-11-01
Contribution to Dexie


dfahlander:
Hi!

After a busy year, I finally more time for Dexie.js again. The Dexie usage and website traffic is growing constantly but it could do better. Microsoft uses Dexie.js in its new replacement of the Todo-app Wunderlist. The release of Dexie.js 2.0 was mentioned in JS Weekly the other month. But there's a lot of work to do and I would appreciate any contributions I can get in the following areas:

* Filing issues and ideas
* Write dexie addons or dexie-based libs.
* Answering and / or fixing issues
* Answerring questions on [stackoverflow (tagged dexie)](https://stackoverflow.com/questions/tagged/dexie?sort=newest)
* Improve docs on [dexie.org](http://dexie.org/docs/) using the Edit button (or clone it locally).
* Implementing [requested features](https://feathub.com/dfahlander/Dexie.js) or your own ideas.

If wanting to know where we're heading, the goals announced earlier this year in [Vision for Dexie](#427) are still valid. If you are about to do a larger contribution, please announce it here as an issue so we can correlate it with other refactoring work.

# Refactoring Dexie

@nponiros, @chrahunt and I are having a discussion on #602 about a much needed refactoring of Dexie.js. I hope to get help doing this, as I will be focused on the addons for the coming month. The goal is to split up Dexie.js into Collection, Table, WhereClause etc and rewrite it in Typescript. Still not clear who will do this work, but I think it needs to be one person that can focus on it fully for a shorter time (max 2 weeks) as it blocks other contributions when being done.

# dexie-observable v2

I have started the process to of a simplification of dexie-observable (v2). Basically, the addon will no longer be the base for dexie-syncable as it solves another use case and can be greatly optimized and simplified if it's only purpose is subscribing to database changes between tabs/windows and not having to supply a base platform for sync. The really cool parts "db.observe()" and "Collection.observe()" in [Vision for Dexie](#427) will be implemented in this addon and it will be based on Rx.Observable.

# dexie-syncable v2

[...]

PR #740, opened 2018-09-08
Update build deps for Dexie Observable

crutchcorn:
As some initial work for Dexie Observable updates (#606 has some mention of what I'm going to be working on here), I went ahead and updated the build deps of Dexie Observable. I have started working on fixing the TS definition but needed the build to work (and figured I'd touch it up while there and break out the commits for each feature/addition) first to test them.

### DIFF
--- a/addons/Dexie.Observable/package-lock.json
+++ b/addons/Dexie.Observable/package-lock.json
@@ -5,12 +5,12 @@
   "requires": true,
   "dependencies": {
     "accepts": {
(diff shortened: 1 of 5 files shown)


dfahlander:
Thanks! 👍 Let's keep this PR open for additional commits for now so we can merge it in when complete.

dfahlander:
Can I merge this in? Seems to be isolated for fixing the build of the addons.

Issue #756, opened 2018-10-22
Invalid type - 'string[]' instead of Table'<any, any>'


StabbarN:
http://dexie.org/docs/Dexie/Dexie.transaction()
About argument `table(s)`:
> Table instances or table names to include in transaction. You may either provide multiple arguments after each other, or you may provide an array of tables. Each argument or array item must be either a Table instance **or a string**.

Supplying strings work but Typescript check says:
> TS2345: Argument of type 'string[]' is not assignable to parameter of type 'Table<any, any>[]'.
  Type 'string' is not assignable to type 'Table<any, any>'.
For example:
```
const stores = ["store1", "store2"]
db.transaction("rw", stores, () => {
    ...
})```

dfahlander:
Solving this issue in master branch so far (which will be Dexie 3.0.0-alpha.x).

PR #757, opened 2018-10-22
Solving issue 756 - able to use string as table args.

dfahlander:
Also correcting the reality that current Transaction is given as argument to the scope function.


dfahlander:
Linking to #756

PR #810, opened 2019-03-22
fallback on Promise

mcshaz:
When building against esNext window.Promise, Node.Promise etc. do not seem to exist, resulting in a runtime error in the code when running on Node.  While this change could be regarded as a hack, it allows Dexie to run.

### DIFF
--- a/src/helpers/promise.js
+++ b/src/helpers/promise.js
@@ -762,7 +762,7 @@ function switchToZone (targetZone, bEnteringZone) {
 }
 
 function snapShot () {
-    var GlobalPromise = _global.Promise;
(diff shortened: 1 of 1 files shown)


dfahlander:
The problem is that Promise is defined locally so it would fallback to Dexie.Promise instead of the native Promise. Could be fixed by renaming function Promise to something else in promise.js. If so, we wouldn't have to pick _global.Promise at all but instead refer to it as just Promise.

mcshaz:
OK - thanks for the quick response.  I'll close this pull request and open an issue. hopefully I get time to get around to it. I will have to have a much deeper look into the differences between a dexie.Promise and the ES native promise.

PR #813, opened 2019-03-22
Fix #812 for IE11

dfahlander:
Old browsers that do not have Promise on window seems to fail. Fixing by verifying typeof Promise === 'object' before applying the #812 workaround.

### DIFF
--- a/src/functions/utils.ts
+++ b/src/functions/utils.ts
@@ -5,7 +5,7 @@ const _global =
     typeof self !== 'undefined' ? self :
     typeof window !== 'undefined' ? window :
     global;
-if (!_global.Promise){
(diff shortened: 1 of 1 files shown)


dfahlander:
@mcshaz, can you review that my addition is ok for esnext compatibility. It turned out to fail when unit tests ran on ie11.

dfahlander:
@mcshaz This PR was faulty as it checked `typeof Promise === 'object'`. Probably this would break the whole idea of the patch as typeof Promise would be 'function' and not 'object' in a new environment. So I've corrected it to `typeof Promise !== 'undefined'`. My fix was to make the code work on Internet Explorer 11 that has no Promise.

If you have a chance to comment on PR #814, that would be great.

PR #814, opened 2019-03-25
Bugfix of previous PR - checking for typeof Promise==='function' instead of 'object'

dfahlander:
typeof Promise is either "undefined" or "function". In #813, it was checked for "object" which would probably never be true (functions used to be "object" in ES3 but then Promise did not exist)

Also added some comments explaining why this statement is there.
#812


mcshaz:
Sorry I took a few days to get back to you - I can confirm the new check:
```if (!_global.Promise && typeof Promise === 'function'){```
works in my ESNext environment

Issue #816, opened 2019-03-29
dbNamesDB on same database as injected IDBFactory


mcshaz:
Just to put a possible feature up for discussion/consideration in the ever more modular js development environment, 

while Dexie allows the class to be instantiated with ```({ indexedDB, IDBKeyRange })``` args, a private instance will have been instantiated for use by the library itself:
```
dbNamesDB // Global database for backing Dexie.getDatabaseNames() on browser without indexedDB.webkitGetDatabaseNames()
```
The conflicting problems are that in its current form, any indexedDB shim (or indeed mock/spy) injected with the instantiation args will be a completely separate database to the internal dbNamesDB, and thus if running in a node environment and using ES6 (asyncronous) modules, "hacks" have to be put in place to make sure the global window object has the shimmed indexedDB added before the Dexie code is encountered.

Of course addressing this problem and querying the dbNamesDB on the same context injected into the instantiator potentially has performance implications. With the current form, the asynchronous operations to establish a transaction and read from the database will be happening while other code is loading, so I guess it is potentially more performant in its current form.

dfahlander:
Thanks! There's also Dexie.dependencies that can be used if not wanting to patch the global object:

```js
var Dexie = require('dexie');
Dexie.dependencies.indexedDB = indexedDB;
Dexie.dependencies.IDBKeyRange = IDBKeyRange;
```

We have that same problem with other static methods, such as [Dexie.delete()](https://dexie.org/docs/Dexie/Dexie.delete()). For those static methods, Dexie.dependencies or global object must have been patched.

dfahlander:
But I could agree that dbNamesDB would be able to use the instance' {indexedDB, IDBKeyRange} instead, since this is something that will be accessed on db.open() when a new DB is created.

mcshaz:
I believe the Dexie dependencies suffer from the same problem i.e. if a transpiler happens to place
1. Dexie module
2. window.indexedDB shim or spy/mock
3. The code: ```Dexie.dependencies.indexedDB = shim_of_indexedDB;```
the getDatabaseNames function will still look at the original window.indexedDB function

PR #619, opened 2017-12-01
Handle security error in Firefox when cookie is disabled

chrmod:
fixes: https://github.com/dfahlander/Dexie.js/issues/617 and https://github.com/dfahlander/Dexie.js/issues/614

we've submited related fixes to PouchDB which were already merged: https://github.com/pouchdb/pouchdb/pull/6890

Issue #617, opened 2017-11-27
SecurityError when storage is disabled (again)


asimen1:
`Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

This was previously fixed in https://github.com/dfahlander/Dexie.js/issues/158 (v1.3.4)

But I am getting this error again (on Chrome 62 when disabling cookies in privacy settings).
Seems like this [commit](https://github.com/dfahlander/Dexie.js/commit/8c4d33b622b03393287291111c932ad09e41d396) removed the try&catch when attempting to access `localStorage`.

Can be easily reproduced with latest Chrome (62) + [fiddle ](https://jsfiddle.net/dfahlander/xf2zrL4p/) + [disabling cookies](https://support.google.com/accounts/answer/61416?co=GENIE.Platform%3DDesktop&hl=en).

(Using latest Dexie release 2.0.1)

dfahlander:
Dexie has removed it's dependency of localStorage so it can't be Dexie that complains.

dfahlander:
I ran the fiddle with cookies disabled and it seems Google Analytics code on jsfiddle was the first issue, and then some other libs that was complaining too.

asimen1:
Maybe fiddle wasn't the best idea for testing this since there are additional errors like you noticed.
However, I clearly see the error from Dexie aswell...

Developer tools image:
![error](https://user-images.githubusercontent.com/12048798/33318790-a6073a84-d444-11e7-96e3-15864ecd1238.png)

This is the line in the code where checks `localStorage` and causes the error: https://github.com/dfahlander/Dexie.js/blob/ffa52d1ad5b39b8316760a50459d4779a6171862/src/Dexie.js#L3187

dfahlander:
I see! Forgot about that one ;)

gregor:
@dfahlander Can this issue get some love with a fix even on the 2.x branch. The Wire app just shows a blank page because of it instead of our cookie warning page. It would be great to get this fixed in order to help our users and decrease the volume of related request to our CS team.

asimen1:
+1
I would also love to see this fixed in 2.x.

gregor:
Thanks for fixing this. Will get this fixed in https//github.com/wireapp/wire-webapp in the morning.

Issue #614, opened 2017-11-14
Security error: Open DB in Firefox 56 and above


01fade:
Hi,

I have built a browser extension using this library to store data locally.
I get an error, when opening the data base with dexie!

```SecurityError: The operation is insecure.```

I have tested the Hello World code and it happens, as well, so I think it's not anything in my code.
My browser extension runs fine in Chrome. 
Would greatly appreciate any help with this! As the Chrome extension has around 78K users and I just launched it for Firefox, it ran fine a couple days ago!

THANKS!

dfahlander:
Anyone else that has had written a Firefox addon that accessed indexedDB? Pinging @ginnipankaj, @achintverma, @damoclark.

01fade:
Hello! I found the root of the problem actually, I totally forgot that I changed my privacy setting in Firefox to "Always use private browsing mode" and that doesn't work with Dexie. So, it was just me, sorry! Closing issue.

Thanks for creating this library btw. I think it's great and makes a lot of things easier, keep up the good work!

dfahlander:
Thanks for sharing.

PR #820, opened 2019-03-30
dbName follows dependencies.indexedDB

mcshaz:
setting Dexie.dependencies.indexedDB = indexedDB;
now updates the databaseNamesEnumerator so that the table of names is in the same database provider

addresses point 1 of issue #816

still has not addressed including the IDBFactory when instantiating a new Dexie object 

all tests passing

### DIFF
--- a/src/classes/dexie/dexie-static-props.ts
+++ b/src/classes/dexie/dexie-static-props.ts
@@ -199,10 +199,18 @@ props(Dexie, {
   // In node.js, however, these properties must be set "manually" before instansiating a new Dexie().
   // For node.js, you need to require indexeddb-js or similar and then set these deps.
   //
-  dependencies: new DexieDependencies({
+  // The try-catch block is because referencing indexedDB in private mode will result in an invalid state error.
+  // it is better to have the failure when dexie open is called.
+  dependencies: new DexieDependencies((() => {
+    try {
+      return {
         indexedDB: (_global.indexedDB || _global.mozIndexedDB || _global.webkitIndexedDB || _global.msIndexedDB) as IDBFactory,
         IDBKeyRange: (_global.IDBKeyRange || _global.webkitIDBKeyRange) as IDBKeyRange
-      }),
+      }
+    } catch {
+      return { indexedDB: null, IDBKeyRange: null }
+    }
+  })()),
 
   // API Version Number: Type Number, make sure to always set a version number that can be comparable correctly. Example: 0.9, 0.91, 0.92, 1.0, 1.01, 1.1, 1.2, 1.21, etc.
   semVer: DEXIE_VERSION,
--- a/src/classes/dexie/dexie-static-props.ts
+++ b/src/classes/dexie/dexie-static-props.ts
@@ -2,7 +2,7 @@ import { Dexie as _Dexie } from './dexie';
 import { props, derive, extend, override, getByKeyPath, setByKeyPath, delByKeyPath, shallowClone, deepClone, getObjectDiff, asap, _global } from '../../functions/utils';
 import { fullNameExceptions } from '../../errors';
 import { DexieConstructor } from '../../public/types/dexie-constructor';
-import { DatabaseEnumerator, databaseEnumerator } from '../../helpers/database-enumerator';
+import { databaseEnumeratorFactory, databaseEnumerator } from '../../helpers/database-enumerator';
 import { PSD } from '../../helpers/promise';
 import { usePSD } from '../../helpers/promise';
 import { newScope } from '../../helpers/promise';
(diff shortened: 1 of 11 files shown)


dfahlander:
Thanks for this PR! I see that tests does not currently pass. Just for your info, I'm currently having issues with travis/karms/browserstack also for the master branch but haven't nailed down the issue yet. Probably due to external causes (changes in the travis or browserstack services).

I will keep trying to find the reason for the travis issues.

mcshaz:
thank you - it was passing until I did something - i'll try and nail it down

dfahlander:
Just want to say, it might be problems with travis/browserstack, since I'm having issues in master branch as well, and that was before your PR.

mcshaz:
can you please let me know when the travis/browserstack issue is fixed, and I'll have another look - thank you.

dfahlander:
Seems the pure dexie tests pass, so that's good. Probably it's only the travis-issue that makes the addon-tests fail. I'll let you know when the travis issues are fixed, but I can't promise when that will be, but as long as it's not fixed, we're blocked on any PRs.

dfahlander:
@mcshaz you can also subscribe to #821 to get notified about the travis-issue.

src/classes/dexie/dexie-static-props.ts
```diff
@@ -198,17 +199,10 @@ props(Dexie, {
   // In node.js, however, these properties must be set "manually" before instansiating a new Dexie().
   // For node.js, you need to require indexeddb-js or similar and then set these deps.
   //
-  dependencies: (()=>{
-    try {
```
dfahlander: The try...catch is important here. We had an issue where Dexie fails completely to initialize when used in private mode. Better if it fails in a controlled way (Dexie still defined but fails in db.open())

mcshaz:
this is just being closed and reopened to rerun the Travis CLI build now that issue is fixed

mcshaz:
passing, and passes when I run tests in IE. Bizarrely, when I first ran the tests in IE, it failed for a single test with:
```
    × modify-with-where(issue-594)
      IE 11.0.0 (Windows 7.0.0)
    There should be 3 Davids
    Expected: 3
    Actual: 2
 ```
but has passed all tests on subsequent occasions. of note the original issue seems to be safari, and I am not sure if the test is truly encapsulated.

to be clear on what this test allows:
```typescript
export class DrugsDBLocal extends Dexie implements IDrugDB {
    constructor(indexedDb?: IDBFactory,
                      dbKeyRange?: typeof IDBKeyRange) {
        if (indexedDb !== void 0) {
            Dexie.dependencies.indexedDB = indexedDb as IDBFactory;
        }
        if (dbKeyRange !== void 0) {
            Dexie.dependencies.IDBKeyRange = dbKeyRange;
        }
        super('DrugData')
```

dfahlander:
Good that the tests succeeds. Internet Explorer is shaky. The browser fail sporadically on the unit tests and always has.

What I found in the review was the try..catch being removed. The reason for the try..catch can be found in issue #614, #617 and PR #619.

mcshaz:
No worries – I’ll put it back in.

mcshaz:
the try catch is back in the original state. having said that, I think it could be potentially improved further - I am as far from an expert on indexedDB as could be, but the error messages & stack trace I have seen are quite unambiguous in Firefox at least - the error is because of insufficient DB permissions. I am not as convinced a null reference would give just as much information. If you want to wait until calling db open, what I would propose is something along the lines of:

```typescript
    public constructor(args?: IDexieDependencies){
        if (args !== void 0){
            this.IDBKeyRange = args.IDBKeyRange;
            this.indexedDB = args.indexedDB;
            this.pIDBError = args.iDBError;
        }
    }
    public get indexedDB(){
        if (this.pIDBError) { throw this.pIDBError}
        return this.pIndexedDB;
    }
... 'dexie-static-props
  dependencies: new DexieDependencies((() => {
    try {
      return {
        indexedDB: (_global.indexedDB || _global.mozIndexedDB || _global.webkitIndexedDB || _global.msIndexedDB) as IDBFactory,
        IDBKeyRange: (_global.IDBKeyRange || _global.webkitIDBKeyRange) as IDBKeyRange
      }
    } catch (err) {
      return { indexedDB: null, IDBKeyRange: null, iDBError: err }
    }
  })()),
```

I have not instituted this as this would be  a breaking change

mcshaz:
just bumping this as so far no response since making the changes a month ago, and wondering if it has been forgotten  - as per your comment, I had added back in the enclosed function & try/catch block to the PR, + added a new comment explaining why the try/catch block. 

On a separate branch (not in this PR) I hold onto the original browser error in order to rethrow when first accessing Dexie

dfahlander:
I haven't got any time for getting a full understanding of the changes. I've had a lot to do, and this PR comes with new logic that is not obvious so i would need to spend more time on it.

mcshaz:
No worries, I had just thought it might have been forgotten about.

dfahlander:
Replaced by #1194

Issue #829, opened 2019-04-11
dexie-export-import has dependency to stream


dfahlander:
As posted on stackoverflow (https://stackoverflow.com/questions/55630801/error-importing-dexie-export-import-in-angular/55634014#55634014) the addon dexie-export-import might not work in certain packaging configs due to a bundled dependency of 'clarinet' that tries to call require('stream') in a try...catch statement.

dfahlander:
Fixed in PR #830 and already released on npm as dexie-export-import@1.0.0-alpha.10.

Keeping this issue open until PR is merged into master.

PR #830, opened 2019-04-11
Remove 'stream' dependency from dexie-export-import

dfahlander:
See #829


dfahlander:
This fix is already released on npm as dexie-export-import@1.0.0-alpha.10.

Don't want to merge this PR until we've fixed the travis CI problems in master first.

PR #863, opened 2019-06-17
Fix #862: Invalid JSON in Filtered Chunked Exports

### LINKED ISSUE #862: dexie-export-import: Invalid JSON formatting in filtered exports
Dexie.js version 2.0.4
dexie-export-import version 1.0.0-beta.11
-> Using default options except for filter

Similar to #854 

With large databases, filtering out large sets of data can result in "ghost commas".

For example, if I apply the following option to my export:
`let exportBlob = await db.export({
    filter: () => false
});`
And if I have a large enough set of data to require a significant amount of chunks, I will find output similar to this for each of my larger tables:
`{"tableName":"testTable1","inbound":true,"rows":[,,,,,,]}`

If I have time to do so (probably not for another day or so), I will submit a PR.

travddm:
- Added unit test for filtered chunked exports (#862)
- Fixed #862:
  - exportAll@dexie/addons/dexie-export-impot/src/export.ts:
    - Code now tracks the number of filtered rows
    - Commas are now appended to table data only when the number of filtered rows in the last iteration is greater than 0
- All tests passing

### DIFF
--- a/addons/dexie-export-import/package-lock.json
+++ b/addons/dexie-export-import/package-lock.json
@@ -973,7 +973,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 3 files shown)


dfahlander:
Thanks a lot!

PR #874, opened 2019-07-18
Fix #872

### LINKED ISSUE #872: dexie-export-import: Out-of-memory Crash On Import
Dexie.js version 2.0.4
dexie-export-import version 1.0.0-beta.12
-> Using default options

While attempting to import a very large (~1.5 GB) previously exported database, chrome repeatedly ran out of memory and crashed. The very same code works fine for smaller files. It is running in a chrome worker, and given a Blob that is properly formatted:

```
const db = await Dexie.import(blob, {
    progressCallback: progressCallback
});
```

After creating an Allocation sampling profile, it became apparent that there was a significantly larger portion of memory being used in comparison to the file size. It rose to ~2GB before crashing.

Here are the most significant profiler results:
image

I'll attempt to fix this and submit a PR.

travddm:
As mentioned in #872, importing large databases had a massive memory overhead, leading to browser crashes.

The reason for this is that the parser couldn't parse objects partially, it had to finish parsing the entire object before making it available to the importer. This caused each import to potentially hold mutliple copies of the entire database in memory.

Changes:
- json-stream now makes partially-processed objects available
  - on creation, objects are given a property: `obj.incomplete = true`
  - when the object is finished being parsed, the `obj.incomplete` property is removed
- import functions fully respect the aforementioned changes

All unit tests pass.

### DIFF
--- a/addons/dexie-export-import/src/import.ts
+++ b/addons/dexie-export-import/src/import.ts
@@ -105,7 +105,7 @@ export async function importInto(db: Dexie, exportedData: Blob | JsonStream<Dexi
           throw new Error(`Primary key differs for table ${tableExport.tableName}. `);
         }
 
-        const sourceRows = tableExport.rows.map(row => row);
+        const sourceRows = tableExport.rows
         
         // Our rows may be partial, so we need to ensure each one is completed before using it
         const rows: any[] = [];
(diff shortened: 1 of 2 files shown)


travddm:
Ah right, I forgot to change that. Thank you for finding it!

dfahlander:
Thanks! I've need it in from my mobile. Hope to release a new package soon when I'm back on the computer

dfahlander:
This PR is released in dexie-export-import@1.0.0-beta.13

PR #912, opened 2019-09-29
Adding array processing to parseIndexSyntax

Seasle:


dfahlander:
This PR would be a significant change in dexie's API that could have effects on existing add-ons that overrides parseIndexSpec. If the purpose is to allow dividing up indexes on different rows, It is possible already using template strings

```js
db.version(x).stores({
  foo: `
    id,
    name,
    age
  `
});
```

PR #917, opened 2019-10-10
Correct type definitions for syncable addon

Jblew:
@oppianmatt and @tunarob from my team found a mistake in type definitions of Dexie.Syncable addon: `disconnect` and `delete` are missing Promise return type.


dfahlander:
Thanks!

PR #953, opened 2019-12-03
Import IDatabaseChange directly in Dexie.Syncable vs re-exporting from api

### LINKED ISSUE #944: Import declaration conflict with TypeScript 3.7
```
ERROR in ../node_modules/dexie-syncable/api.d.ts:31:9 - error TS2440: Import declaration conflicts with local declaration of 'IDatabaseChange'.
31 import {IDatabaseChange} from 'dexie-observable/api';
```

Seems TypeScript made a potentially (in this case definitely) breaking change in 3.7:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#local-and-imported-type-declarations-now-conflict

mloffer:
Fixes #944. Note that the [npm page](https://www.npmjs.com/package/dexie-syncable) contains the following snippet:

```
import Dexie from 'dexie';
import 'dexie-syncable'; // will import dexie-observable as well.
```

I am unclear on whether consumers need access to IDatabaseChange, but if so, dexie-observable will also need to be imported.

### DIFF
--- a/addons/Dexie.Syncable/api.d.ts
+++ b/addons/Dexie.Syncable/api.d.ts
@@ -17,19 +17,18 @@
  *
  * By separating module 'dexie-syncable' from 'dexie-syncable/api' we
  * distinguish that:
- * 
+ *
  *   import {...} from 'dexie-syncable/api' is only for getting access to its
  *                                          interfaces and has no side-effects.
  *                                          Typescript-only import.
- * 
+ *
  *   import 'dexie-syncable' is only for side effects - to extend Dexie with
  *                           functionality of dexie-syncable.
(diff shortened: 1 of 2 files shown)


dfahlander:
Thanks!

oppianmatt:
so I ran into this problem with the new typescript which I need for other purposes. I saw the issue has been fixed, could it also be published to the npm packages version please?

I tried upgrading to 3.0.0-rc1 in npm, but typescript had more errors from trying to use dexie-observable and dexie-syncable (they had no next relesases in npm). It is possible to use/try the 3.0.0-rc1 release with dexie-syncable?

PR #955, opened 2019-12-17
typeson/typeson-registry update

brettz9:
- Refactoring: With typeson/tyepson-registry update, can avoid `finalize` patching
- npm: Update typeson/typeson-registry

If you like, I can also tack on to this PR:

1. Updating the other devDeps
2. Remove your internal use of `arraybuffer` as it is already present with the typeson-registry structured cloning preset, and with basically the same implementation but relying on the defaulting of `base64-arraybuffer-es6` and caching existent buffers for the sake of cyclical references.
3. Replace your internal use of `tson-typed-array` with `typeson-registry`'s `typed-arrays`. (While the `typed-arrays-socketio` type works with your code as it is if you prefer, the `socketio` preset does not (despite `socketio` now passing `typeson-registry` tests); I haven't looked closely enough at your code to see which makes more sense.)

(By the way, we don't have the preservation of `byteOffset` and `length` in `typeson-registry`'s `typed-arrays-socketio` type as we do with the `typed-arrays` type. While we could preserve it, since we are just passing on the buffer as is unless the type is detected as `ArrayBuffer`, I guess maybe it doesn't fit there.)

If this is too much info to review now, feel free to just review the PR as is.

### DIFF
--- a/addons/dexie-export-import/package-lock.json
+++ b/addons/dexie-export-import/package-lock.json
@@ -33,9 +33,9 @@
       "dev": true
     },
     "agent-base": {
-      "version": "4.2.1",
-      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.2.1.tgz",
-      "integrity": "sha512-JVwXMr9nHYTUXsBFKUqhJwvlcYU/blreOEUkhNR2eXZIvwd+c+o5V4MgDPKWnMS/56awN3TRzIP+KoPn+roQtg==",
+      "version": "4.3.0",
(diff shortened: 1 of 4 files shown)


dfahlander:
Thanks! I will have a look into it. If merging this I might get some issues with backward compatibility though, I suppose. Will probably need to bump major version or find some other solution to people that have already exported files before this change... Anyway, thanks again for showing how the new version of typeson-registry has fixed the issue that I was working around 😍

dfahlander:
Closing this for housekeeping. Right now the impact of using it would be backward incompatibility and it's not worth it when we already have a suitable workaround. It's nice though that we have this PR so we can use it as a reference for future work.

Issue #942, opened 2019-11-25
Upgrade function not running on DB upgrade


joseprio:
We noticed some of our users were missing an index that was added through an upgrade. I decided to make a little experiment to see what would happen if the browser "dies" in the middle of an upgrade by using a long `Dexie.waitFor`, and the upgrade function was not called the second time, even though the DB upgrade process was resumed.

I wanted to create a unit test that reproduces this error, but unfortunately I couldn't find a way to express something that doesn't fail nor succeed.

Is this a known problem?

dfahlander:
It's not a known problem, no. If an upgrade transaction fails, it will not upgrade the version, no matter if you were using Dexie.waitFor() or not. However, make sure you return the promise from your upgrader. In case you do not use that, and the Dexie.waitFor() times out (fails with TimeoutError) or stumbles on another error, and that failing promise must be returned to the caller so that it can abort the upgrade transaction. This would be automatically handled if your upgrader is an async function. Otherwise, make sure to return the whole promise flow.

The only known issue regarding upgrades, is that if the initial, database-creating transaction fails, IndexedDB will still create the DB at version 0. This scenario is worked around in Dexie by deleting the database when that happens.

The most common pitfall with Dexie upgrades, is if the version that lacks the index was published at some point, and then adding an index to the same version without incrementing the version number. Dexie won't complain if the installed DB has a different set of indexes than the ones declared.

joseprio:
@dfahlander thanks a lot for your quick reply.

I tweaked one of your jsfiddle, and I can reproduce the problem in Firefox 48 and 58, but not in Firefox 69 or newest Chrome, using Dexie 2.0.4 .
```
function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}
db.version(1).stores({
	friends: '++id,name,age'
});
db.version(2).upgrade(tx => Dexie.waitFor(sleep(800000)));
```
Could it be a browser bug?

dfahlander:
Was the result that the DB was on version 2 in the failed results? If so, you might have nailed down the issue. But I would like to nail it down further. 

* If you provide a shorter timeout than the default of 60 seconds to [Dexie.waitFor()](https://dexie.org/docs/Dexie/Dexie.waitFor()), and pass a lesser value to the sleep() function, is it then also failing on Firefox 48 or 58?

* If you provide an alternate set of indexes in version 2 and reproduce the problem, what would be the set of indexes and version number present after reproducing it?

[...]

PR #959, opened 2019-12-26
Base schema upgrade on current schema information from the database

joseprio:
Mitigates the problem described in #942

### DIFF
--- a/src/classes/dexie/dexie-open.ts
+++ b/src/classes/dexie/dexie-open.ts
@@ -65,6 +65,7 @@ export function dexieOpen (db: Dexie) {
           } else {
               upgradeTransaction.onerror = eventRejectHandler(reject);
               var oldVer = e.oldVersion > Math.pow(2, 62) ? 0 : e.oldVersion; // Safari 8 fix.
+              db.idbdb = req.result;
               runUpgraders(db, oldVer / 10, upgradeTransaction, reject);
           }
       }, reject);
--- a/src/classes/version/schema-helpers.ts
+++ b/src/classes/version/schema-helpers.ts
@@ -77,13 +77,10 @@ export function updateTablesAndIndexes(
   // Each transaction object will contain the table set that was current in that version (but also not-yet-deleted tables from its previous version)
   const queue: UpgradeQueueItem[] = [];
   const versions = db._versions;
-  const oldVersionStruct = versions.filter(version => version._cfg.version === oldVersion)[0];
-  if (!oldVersionStruct) throw new exceptions.Upgrade(
-    "Dexie specification of currently installed DB version is missing");
-  let globalSchema = db._dbSchema = oldVersionStruct._cfg.dbschema;
+  let globalSchema = db._dbSchema = buildGlobalSchema(db, db.idbdb, idbUpgradeTrans);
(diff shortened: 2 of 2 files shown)


dfahlander:
This looks like an elegant solution. Thanks a lot! I've briefly reviewed the changes but want to do it more thoroughly before merging it in.

timbru31:
Ouch! This PR breaks the update mechanism completely :(
Since the `buildGlobalSchema` function is used, the `unique (&)` flag is lost, because `createIndexSpec` (see line 300) always passes false.

This results in a diff like this:

Screenshot 2020-02-25 at 18 10 40

Therefore, the `recreate` flag will be set and Dexie stops with the `UpgradeError` error:
Screenshot 2020-02-25 at 18 09 58


Before this PR, the old schema is used (via `oldVersionStruct._cfg.dbschema;`) and the `recreate` flag is `false`:
Screenshot 2020-02-25 at 18 16 55


The diff between our two version is the following:

```ts
		versions.push({
			versionNumber: 2,
			stores: {
				walksheet: '++id',
				order: '&uuid,created,modified,status',
				buildings: '&geb_id,adresse,strasse',
				people: '&kundenr,vorname,nachname'
			}
		});
		versions.push({
			versionNumber: 3,
			stores: {
				walksheet: '++id,&plz,lastImport',
				order: '&uuid,created,modified,status',
				buildings: '&geb_id,adresse,strasse,plz',
				people: '&kundenr,vorname,nachname'
			}
		});
```


As you can see, it should only add new indices.
Please correct me if I'm wrong, but IMO this is a major regression.

joseprio:
This only applies for the primary key; is it possible the table was created
manually?

timbru31:
No every interaction with the IndexedDB is made through Dexie. Tables were created by Dexie v2 though.

joseprio:
Ok, I think that the problem here is that uuid being the primary key makes
it inherently unique, and therefore that ampersand has no effect.

In a way, I’d argue that declaring it unique is incorrect, as unique
indices are not the same as primary keys.

timbru31:
Indeed the docs state `Detail: Primary keys are implicitly marked as unique.` - given the fact that it worked quite well since years, I'd argue that explicitly marking a primary key as unique should not cause any issues (as in our case).  

I'll debug tomorrow if what happens if I remove the explicit uniqueness (&)

dfahlander:
It's ok to declare primary key as being unique. I've seen people do that in order to be more explicit. I think we must account for that in the upgrade function and not regard the primary key as changed in case the uniqueness differs, as primary keys are unique not matter how they were declared. I suppose it is an easy fix in schema-helpers.ts when checking if the primary key differs or not.

We should create a repro of this issue in test-upgrade.js and fix before releasing 3.0.0.

dfahlander:
I will not have the time this week so if anyone would have the time, that would be perfect. Otherwise I try fix it next week.

timbru31:
As expected, the migration runs successfully without the explicit unique modifier.

I've create a WIP PR here that adds a, currently, failing test:https://github.com/dfahlander/Dexie.js/pull/984

Hopefully I'll have enough time tomorrow or friday to come up with a fix for the failing test ;)

PR #962, opened 2019-12-29
fix the compiler warnings & stuff

heavyk:
I started out attempting to make an es6 version of dexie that does not have all of the backwards compatible browser stuff in it. my target platform is chrome/electron, and considering the new IE is based on chromium now. I also want to have a version have a version that is also lighter as well (right now dexie is 60KB+ minimised) so I can include only the parts that I need.

anyway, this is my first time using typescript, and honestly it's quite nice :) results that the compiler spits out some warnings and there are tons of unused variables and imports which my editor complains about as well. I fixed all of them. unfortunately, my editor also deletes trailing whitespace. I noticed in your latest commit, trailing whitespace was also trimmed so I feel less bad about it.

just in case you want it, this seems to be most of the major fixes.

---

in addition, I was wondering if it's possible to (on chrome/electron) use the native unpatched promises? do you know of any major things that can go wrong if I change dexie use those? (the reason why I say is because you have added things like long stack trace support, which chrome supports natively and other things.) that'll probably be the next thing I try

### DIFF
--- a/addons/Dexie.Observable/src/Dexie.Observable.js
+++ b/addons/Dexie.Observable/src/Dexie.Observable.js
@@ -55,7 +55,7 @@ var override = Dexie.override;
 var Promise = Dexie.Promise;
 var browserIsShuttingDown = false;
 
-export default function Observable(db) {
+export default function Observable(db: Dexie) {
     /// <summary>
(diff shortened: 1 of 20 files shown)


dfahlander:
Closing for housekeeping

Issue #969, opened 2020-01-23
[suggestion / discussion] Bulk creation methods to return all results


PVermeer:
Hey dfahlander,

First thanks for creating this amazing library. I find it well written and easy to use / expand.

The bulk creation methods currently only return the last result. I've created some plugins that have custom id generation methods and that add immutability to the input objects. When bulk creating records I keep running into the issue that I need all results (id's) to update state correctly.

Immutability is not something I would like to give up on.

A little research into the source revealed that I probably could write another plugin that overrides these methods but in this case I feel like it should be a native option.

**What I found:**
The mutate function is already returning all results:
https://github.com/dfahlander/Dexie.js/blob/08f9138cd7f79b5a725a7f2046e1f6bfc5fb8d11/src/dbcore/dbcore-indexeddb.ts#L181-L190

The bulkAdd method is just not using it:
https://github.com/dfahlander/Dexie.js/blob/08f9138cd7f79b5a725a7f2046e1f6bfc5fb8d11/src/classes/table/table.ts#L381-L398

**Suggestion**
_Default return all results_(breaking change):
Since there is an upcoming major release it could be changed to return al results as a new default.

_Optional options object_(non-breaking change):
`bulkAdd(someRecords, { allResults: true })`

This is all based on a quick look through the source.

I would like your opinion about this. Also I could make PR if you like with the proposed and tested changes.

Thanks

dfahlander:
Thanks! The original reason for bulkAdd() not returning all results, is that it allows bulkAdd to be more optimized since it does not have to listen to onsuccess event. This is the key performance boost for the bulk methods. However, in the DBCore rewrite, I found a middle-way in dbcore-indexeddb, to know which requests that would need to listen for onsuccess. Only for those objects that are not provided with a key will need to do that, because we could already know the result from the request's keys otherwise. So, yes, DBCore has the interface needed and dbcore-indexeddb implements it.

I did not come to a conclusion on how to fix the issue with the breaking change. But I suppose we could have that breaking change in 3.0. It would only regard to autoincremented keys that are not already provided in request.

We're on release-candidate now and I am not sure if a breaking change is ok at this point though. I would like bulkAdd() to return array and not have the options, but maybe we'd have to go for the option solution at this point.

Please do a PR if you have the time. It would be very nice with a unit test also somewhere in [tests-table.js](https://github.com/dfahlander/Dexie.js/blob/master/test/tests-table.js).

PVermeer:
[...]

PR #973, opened 2020-01-27
Feature:  optionally all keys are returned on creating methods

PVermeer:
See Issue: #969 

**Added code to support an options object on bulkAdd() and bulkPut() methods.**

These methods now accept an optional object on the second or third parameter. Instead of receiving only the last key it is now possible to use the, already implemented feature, to return all keys via { allKeys: true }:

_Example:_
`bulkAdd(someObject[], { allKeys: true })`
or
`bulkAdd(someObject[], someKey[], { allKeys: true })`

**Typings**
Also included improved typings for these methods. Via overloads, the correct return type (TKey | TKey[]) will be used.

**Tests**
Added test for the updated methods and some tests for backward compatibility (**no breaking changes**).

**Performance**
Did not add any performance tests (didn't know how and where you wanted them). I did some testing in my own app. Could not find any meaningful / significant differences. So seems ok.

**Documentation**
Not updated because different repo.

### DIFF
--- a/src/public/types/table.d.ts
+++ b/src/public/types/table.d.ts
@@ -47,7 +47,14 @@ export interface Table<T=any, TKey=IndexableType> {
   delete(key: TKey): PromiseExtended<void>;
   clear(): PromiseExtended<void>;
   bulkGet(keys: TKey[]): PromiseExtended<T[]>;
-  bulkAdd(items: T[], keys?: IndexableTypeArrayReadonly): PromiseExtended<TKey>;
-  bulkPut(items: T[], keys?: IndexableTypeArrayReadonly): PromiseExtended<TKey>;
-  bulkDelete(keys: IndexableTypeArrayReadonly) : PromiseExtended<void>;
+
+  bulkAdd<B extends boolean>(items: T[], keys: IndexableTypeArrayReadonly, options: { allKeys: B }): B extends true ? Promise<TKey[]> : Promise<TKey>;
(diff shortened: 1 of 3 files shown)


src/public/types/table.d.ts
```diff
@@ -47,7 +47,14 @@ export interface Table<T=any, TKey=IndexableType> {
   delete(key: TKey): PromiseExtended<void>;
   clear(): PromiseExtended<void>;
   bulkGet(keys: TKey[]): PromiseExtended<T[]>;
-  bulkAdd(items: T[], keys?: IndexableTypeArrayReadonly): PromiseExtended<TKey>;
-  bulkPut(items: T[], keys?: IndexableTypeArrayReadonly): PromiseExtended<TKey>;
-  bulkDelete(keys: IndexableTypeArrayReadonly) : PromiseExtended<void>;
+
+  bulkAdd<B extends boolean>(items: T[], keys: IndexableTypeArrayReadonly, options: { allKeys: B }): B extends true ? Promise<TKey[]> : Promise<TKey>;
+  bulkAdd<B extends boolean>(items: T[], options: { allKeys: B }): B extends true ? Promise<TKey[]> : Promise<TKey>;
```
timbru31, line 52: Shouldn't the return type be `PromiseExtended` instead of a plain `Promise`? (applies to all new bulkAdd and bulkPut overloads)

PVermeer: Whoops, I think you are right.

PVermeer: Added fix in PR #980

PR #984, opened 2020-02-26
WIP: Correctly handle explicit unique modifiers for primary keys in migration

timbru31:
Step 1: adds test case for migration with explicit unique primary key
Step 2: fix the migration code (TODO)

See https://github.com/dfahlander/Dexie.js/pull/959#issuecomment-590973310 and following comments for more information.


dfahlander:
Thanks for the repro. I continued this PR in #1005 so we can close this one.

PR #1005, opened 2020-03-24
Timbru31 fix/explicit unique primary key

dfahlander:
This PR is based on #984 and completes it.


timbru31:
great, thanks for continuing my PR!

Issue #802, opened 2019-02-24
`IUpdateChange` missing `obj` and `oldObj`


madeken:
I think [IUpdateChange](https://github.com/dfahlander/Dexie.js/blob/e956de3b74cf3ac5f1ed2fa2b97b53e4fe60a3e1/addons/Dexie.Observable/api.d.ts#L31) is missing  `obj` and `oldObj`  (at least when I console.log the actual event it contains that)

moonmeister:
Yes, this is accurate best I can tell.

PR #1020, opened 2020-04-16
fix: add missing properties on observable change

moonmeister:
## Summary
Adds missing `obj` and `oldObj` properties to observable change types and update tests to pass.

## Notes
I might be missing something but it seems the interfaces are. If anything needs reworking let me know I'm happy to make the changes. 

## Related
fixes: #802


moonmeister:
hmm, I fixed some of the larger formatting issues after my prettier defaults changed things. I can fix all of them if needed. seems like maybe it'd be good to add a formatter to the project (happy to do a pr).

moonmeister:
looks like this has been fixed in a different pr

Issue #791, opened 2019-01-17
Dexie.Observable: Cannot read property 'indexOf' of null


bennyn:
We are using `dexie-observable` in our web application and our error reporting shows the following error for some of our clients:

> Message: Cannot read property 'indexOf' of null

It complains about [the following code](https://github.com/dfahlander/Dexie.js/blob/v2.0.4/addons/Dexie.Observable/src/on-storage.js#L6):

> if (event.key.indexOf("Dexie.Observable/") === 0) 

So it seems like that the `event` does not always has a `key` property, so the `indexOf` operation fails. Can you double check that?

dfahlander:
It is reacting to the "storage" event. When storage is cleared, it may not come with a key. A valid fix would be to null-check event.key.

PR #1040, opened 2020-05-05
null check key first

disarticulate:
Dexie.Observable: Cannot read property 'indexOf' of null #791

https://github.com/dfahlander/Dexie.js/issues/791

>It is reacting to the "storage" event. When storage is cleared, it may not come with a key. A valid fix would be to null-check event.key.


dfahlander:
Thanks!

PR #1062, opened 2020-06-12
Adds missing fields to Dexie-Observable TS type definitions

### LINKED ISSUE #1057: Dexie-Observable Typescript API missing fields
The Typescript declaration for the `Dexie.Observable.DatabaseChange` appear to be incomplete. In particular, the `IUpdateChange` interface is missing `obj` and `oldObj`, and the `IDeleteChange` interface is missing `oldObj`, which prevents Typescript users from being able to tell which object was changed.

JS Implementation:
https://github.com/dfahlander/Dexie.js/blob/v3.0.1/addons/Dexie.Observable/src/Dexie.Observable.js#L42-L51

TS declaration:
https://github.com/dfahlander/Dexie.js/blob/v3.0.1/addons/Dexie.Observable/api.d.ts#L24-L44

Is this on purpose, or is this a mistake? If the latter, I'd be happy to submit a PR to fix this.

Thank you for maintaining Dexie — it makes IndexedDB actually sane to use, and I can't imagine working with IndexedDB without it.

calvinwyoung:
This PR resolves #1057 by adding a few missing fields to the TS type definitions for the Dexie-Observable API.

The [Dexie documentat](https://dexie.org/docs/Observable/Dexie.Observable.DatabaseChange) declares that `DatabaseChange` objects should have `obj` and `oldObj` depending on the change type, as well as an optional `source` field. These fields are missing from the current Typescript definitions:
https://github.com/dfahlander/Dexie.js/blob/v3.0.1/addons/Dexie.Observable/api.d.ts

### DIFF
--- a/addons/Dexie.Observable/api.d.ts
+++ b/addons/Dexie.Observable/api.d.ts
@@ -1,44 +1,50 @@
 /**
  * API for Dexie.Observable.
- * 
+ *
  * Contains interfaces used by dexie-observable.
- * 
+ *
  * By separating module 'dexie-observable' from 'dexie-observable/api' we
  * distinguish that:
- * 
+ *
  *   import {...} from 'dexie-observable/api' is only for getting access to its
  *                                            interfaces and has no side-effects.
  *                                            Typescript-only import.
- * 
+ *
  *   import 'dexie-observable' is only for side effects - to extend Dexie with
  *                             functionality of dexie-observable.
  *                             Javascript / Typescript import.
- * 
+ *
  */
 export const enum DatabaseChangeType {
     Create = 1,
     Update = 2,
-    Delete = 3
+    Delete = 3,
 }
 
 export interface ICreateChange {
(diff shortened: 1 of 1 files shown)


juzhiyuan:
How about `db.on("changes")`?

calvinwyoung:
> How about `db.on("changes")`?

Hmm interesting, it looks like Dexie-Observable *should* be handling that already, but I'm not sure why VSCode isn't catching that for me:
https://github.com/dfahlander/Dexie.js/blob/master/addons/Dexie.Observable/src/Dexie.Observable.d.ts#L43-L47

juzhiyuan:
Yes, VSCode just highlights some errors when listening on changes event...

dfahlander:
There are some rolling issues after going over from dexie 2 to dexie 3. Dexie is defined differently in dexie 3 and we should recap how we have to change augmenting of dexie.

PR #1081, opened 2020-07-19
Run reading hooks on Table.bulkGet

### LINKED ISSUE #1079: mapToClass not applied to bulkGet results
Table.mapToClass is not applied to the results of Table.bulkGet. (Using Dexie 3.0.1)

Example: https://codesandbox.io/s/dexie-maptoclass-bulkget-test-7x15x?file=/src/index.ts

I think the underlying issue is that [mapToClass relies on the readHook](https://github.com/dfahlander/Dexie.js/blob/v3.0.1/src/classes/table/table.ts#L248-L267), and that hook is run in [get](https://github.com/dfahlander/Dexie.js/blob/v3.0.1/src/classes/table/table.ts#L80) but not in [bulkGet](https://github.com/dfahlander/Dexie.js/blob/v3.0.1/src/classes/table/table.ts#L363-L370).

It's not entirely clear from the [mapToClass docs](https://dexie.org/docs/Table/Table.mapToClass()) whether this is expected, but it seems odd. (Especially since it looks like all the other bulk operations *do* run their corresponding CRUD hooks via the hooksMiddleware.)

medmunds:
Fire the reading hooks on the results of Table.bulkGet, as in Table.get.

This fixes the problem where Table.bulkGet results don't match Table.get results when Table.mapToClass is in effect. (mapToClass leverages the reading hook.)

Fixes #1079

### DIFF
--- a/src/classes/table/table.ts
+++ b/src/classes/table/table.ts
@@ -365,7 +365,7 @@ export class Table implements ITable<any, IndexableType> {
       return this.core.getMany({
         keys,
         trans
-      });
+      }).then(result => result.map(res => this.hook.reading.fire(res)));
     });
   }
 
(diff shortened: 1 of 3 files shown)


dfahlander:
Thanks a lot! 👍

PR #1098, opened 2020-08-06
Fix Table.bulkGet return typing to include undefined

medmunds:
This improves the typing for Table.bulkGet() to indicate
the returned array may include undefined items.

(Per the [bulkGet docs][], "For those keys that do not
exist in the database, undefined will be returned in their
place.")

[bulkGet docs]: https://dexie.org/docs/Table/Table.bulkGet()


medmunds:
(Not really sure how to test this, short of pulling in dtslint or tsd. But I did add a `bulkGet` to test-typings.ts, with a commented out line that should cause an error since it doesn't handle undefined.)

medmunds:
(Build passed; not sure why GH checks aren't updating: https://travis-ci.org/github/dfahlander/Dexie.js/builds/715647161)

dfahlander:
Thanks a lot!

PR #1106, opened 2020-08-25
#1105 Allow readonly arrays for bulk put & get methods

### LINKED ISSUE #1105: bulkAdd & bulkPut typing: accept a `readonly` array
We follow immutability practices in our codebase, but because of the current typings, `bulkAdd` and `bulkPut` do not accept a `readonly TEntity[]`, only `TEntity[]`.

If I understand it correctly, the typing could be safely relaxed to allow both the readonly and non-readonly versions of the array.

This would be a small change and would avoid a type assertion in the consuming code.

Thank you!

wh1t3cAt1k:
Closes #1105 

@dfahlander You mentioned that it's going to be a tsc error, but seemingly the build and tests all pass without any errors.

### DIFF
--- a/src/public/types/table.d.ts
+++ b/src/public/types/table.d.ts
@@ -48,13 +48,13 @@ export interface Table<T=any, TKey=IndexableType> {
   clear(): PromiseExtended<void>;
   bulkGet(keys: TKey[]): PromiseExtended<(T | undefined)[]>;
 
-  bulkAdd<B extends boolean>(items: (T[] | readonly T[]), keys: IndexableTypeArrayReadonly, options: { allKeys: B }): PromiseExtended<B extends true ? TKey[] : TKey>;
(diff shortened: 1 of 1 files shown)


dfahlander:
Thanks. Yes all tests pass for this PR: https://travis-ci.org/github/dfahlander/Dexie.js/builds/720915693.

Question: Do we need to use `(T[] | readonly T[])`. Could we just use `readonly T[]`. A normal array would also be assignable to a readonly array.

wh1t3cAt1k:
@dfahlander good catch, indeed - fixed.

wh1t3cAt1k:
@dfahlander For some reason, the check seems to be stuck... Is it normal or should the CI be relaunched?

dfahlander:
Thanks for the fix! Yes, I don't know why the github/travis integration seem to be faulty. The build went fine anyway https://travis-ci.org/github/dfahlander/Dexie.js/builds/720938812 so I will merge.

PR #1115, opened 2020-09-08
use error keys insted of values for typescript

rtm7777:
Hello,

I tried to check dexie error name in my typescript project next way:
```
  } catch (err) {
    switch (err.name) {
      case Dexie.errnames.Constraint:
        console.log('already exist')
        break
      default:
        console.log('other error')
    }
  }
```
And I have the next error:
```
Property 'Constraint' does not exist on type '{ OpenFailedError: "OpenFailedError"; VersionChangeError: "VersionChangeError"; SchemaError: "SchemaError"; UpgradeError: "UpgradeError"; InvalidTableError: "InvalidTableError"; ... 24 more ...; BulkError: "BulkError"; }'.
```

I think it's because you use values of error names as keys

But I'm new to TS so I may be wrong

Thanks!

### DIFF
--- a/src/errors/errors.d.ts
+++ b/src/errors/errors.d.ts
@@ -1,6 +1,7 @@
 import {
   DexieError,
   DexieErrorConstructor,
+  DexieErrors,
   ModifyError,
   ModifyErrorConstructor,
   BulkError,
@@ -17,6 +18,6 @@ export declare const exceptions : ExceptionAliasSet & {
   Type: ErrorConstructor,
   Range: ErrorConstructor
 };
-export declare const errnames : {[P in keyof ExceptionSet]: P};
+export declare const errnames : DexieErrors;
 export declare const fullNameExceptions : ExceptionSet;
 export declare function mapError (domError, message?) : DexieError;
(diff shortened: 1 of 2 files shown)


src/errors/errors.d.ts
```diff
@@ -17,6 +18,6 @@ export declare const exceptions : ExceptionAliasSet & {
   Type: ErrorConstructor,
   Range: ErrorConstructor
 };
-export declare const errnames : {[P in keyof ExceptionSet]: P};
+export declare const errnames : {[P in keyof DexieErrors]: P};
```
dfahlander: I think the correct typing here should be just:
```ts
export declare const errnames: DexieErrors;
```
...so that we get {Modify: "ModifyError"} rather than {Modify: "Modify"}.

src/public/types/dexie-constructor.d.ts
```diff
@@ -50,5 +50,5 @@ export interface DexieConstructor extends DexieExceptionClasses {
   //IndexSpec: {new():IndexSpec}; //? Deprecate
   Events: (ctx?: any)=>DexieEventSet;
 
-  errnames: {[P in keyof DexieExceptionClasses]: P};
+  errnames: {[P in keyof DexieErrors]: P};
```
dfahlander: Same here:
```ts
errnames: DexieErrors;
```

dfahlander (changes requested):
Thanks for finding this typing issue! Do you agree on my comments? If so, could you commit the 2 changes and push it back?

rtm7777:
@dfahlander thanks for taking a look at my PR, changes pushed

dfahlander (approved):

Issue #1070, opened 2020-07-05
Filtered Export produces invalid JSON file with an extra comma in data row array  


alperyazgan:
Hi,

Today I faced a problem with filtered export json file;
Normally I do filtering for many tables without any problems, but this table strangely produces an extra comma in the exported data row array;

Here is the main export function I use for the table;

```
async function exportMember(memberNo, exportName) {
    if (db === null || memberNo === "") return false;

    memberNo = parseInt(memberNo);

    let rowCount = 0;
    let dbAsBlob = await db.export({
      prettyJson       : true,
      progressCallback : exportProgress,
      noTransaction    : true,
      filter           : (table, value) => {
        let returnValue = table === "member" && value.memberNo === memberNo;
        if (returnValue) rowCount++;
        return returnValue;
      }
    });
    //let jsonDB = JSON.parse(await dbAsBlob.text()); // It has a extra "," in the row list

    //jsonDB.data.tables.forEach(function (table) {
    //  table.rowCount = table.name === "member" ? rowCount : 0;
    //});
    //dbAsBlob = new Blob([JSON.stringify(jsonDB, undefined, 2)], { type : dbAsBlob.type });
    download(dbAsBlob, `${exportName || "ExportedMemberWithNo("}${memberNo}).json`, "application/json");

    return true;
  }
```

and this func generates a file like this

```
{
  "formatName": "dexie",
  "formatVersion": 1,
  "data": {
    "databaseName": "MemberDB",
    "databaseVersion": 1,
    "tables": [{ "name": "member", "schema": "++id,&memberNo", "rowCount": 1 }],
    "data": [
      {
        "tableName": "member",
        "inbound": true,
        "rows": [
          {
            "memberNo": 2592,
            "fullName": "Some Name",
            "photo": "data:image/jpeg;base64, very long jpeg data",
            "eMail": "mail@hotmail.com",
            "mobilePhoneNo": "5555555",
            "bloodType": "A Rh(+)",
            "id": 860,
            "birthday": "01.03.1978",
            "someFieldsAgain": "with some data",
            "workPhoneNo": null
          }
          ,
        ]
      }
    ]
  }
}
```

I use this approach for the other tables with no problem, only difference is the total row count which is 5500, the other tables have less than 100 rows.

dfahlander:
Thanks!

I think this issue is a "good first issue" for any contributor to fix.

1. Fork or update your existing fork of Dexie.js, npm install.
[...]

PR #1125, opened 2020-09-23
Fix export

senadglojnaric:
Hello, here is my Pull Request for fixing an issue #1070 . I also added a test in edge-cases.

### DIFF
--- a/addons/dexie-export-import/src/export.ts
+++ b/addons/dexie-export-import/src/export.ts
@@ -85,7 +85,13 @@ export async function exportDB(db: Dexie, options?: ExportOptions): Promise<Blob
       const table = db.table(tableName);
       const {primKey} = table.schema;
       const inbound = !!primKey.keyPath;
-      const LIMIT = options!.numRowsPerChunk || DEFAULT_ROWS_PER_CHUNK;
+
+      //issue #1070 //////////////////////////////////
+      const rowsPerChunk= options!.numRowsPerChunk || DEFAULT_ROWS_PER_CHUNK; 
+      const rowCount = await table.count(count => count);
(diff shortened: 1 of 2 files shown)


dfahlander (commented):
I don't see how this fixes the issue. The unit test has to reproduce the error. Does the unit test fail if you remove your fix?

addons/dexie-export-import/src/export.ts
```diff
@@ -85,7 +85,13 @@ export async function exportDB(db: Dexie, options?: ExportOptions): Promise<Blob
       const table = db.table(tableName);
       const {primKey} = table.schema;
       const inbound = !!primKey.keyPath;
-      const LIMIT = options!.numRowsPerChunk || DEFAULT_ROWS_PER_CHUNK;
+
+      //issue #1070 //////////////////////////////////
+      const rowsPerChunk= options!.numRowsPerChunk || DEFAULT_ROWS_PER_CHUNK; 
+      const rowCount = await table.count(count => count);
```
dfahlander, line 91: Why is this needed?

senadglojnaric:
Yes, the unit test fails if the fix is removed.

addons/dexie-export-import/src/export.ts
ghost: Extra comma appears only when the numRowsPerChunk is smaller then the number of rows in the table, wether is defined by the user or a default value.

dfahlander:
But you are not solving the problem. The reason for having the option rowsPerChunk is to save memory. The bug is not solved by working around the issue and increasing the limit. The whole idea is that we want to save memory here, so that databases with a million rows can be exported in chunks. The bug is that this fails if there's a filter function.

I close this for now.

PR #1129, opened 2020-09-25
Cleanup: ensure db.folks update hook is removed after unit test completion

jayaddison:
This is a small unit test cleanup discovered during preparation of an unrelated pull request.

One of the `Table` unit tests uses an `updating` hook to make assertions on the expected behaviour, but does not remove the hook after the unit test completes.

This cleanup ensures that the hook is removed, so that it doesn't run and apply assertions for any subsequent/unrelated unit tests.


dfahlander (approved):
Great you found this, thanks!

PR #1138, opened 2020-10-06
Dexie.Observable: startObserving function: remove read-only query in order to avoid TransactionInactiveError

### LINKED ISSUE #1126: OpenFailedError: TransactionInactiveError when using indexeddbshim + dexie
I'm adding some database-related [unit test coverage](https://github.com/openculinary/frontend/blob/6ed2bb659970caa2cfb027d8fde4b28d05c2bc79/src/app/database.spec.ts#L7-L9) for a Dexie-powered browser application written in JavaScript+TypeScript.

Unit tests are [executed via `mochapack`](https://github.com/openculinary/frontend/blob/6ed2bb659970caa2cfb027d8fde4b28d05c2bc79/Makefile#L34-L35) (`webpack` to compile the app followed by `mocha` test execution), and to provide the IndexedDB API methods, an in-memory [`indexeddbshim` instance is injected](https://github.com/openculinary/frontend/blob/6ed2bb659970caa2cfb027d8fde4b28d05c2bc79/setup.js#L11) at test runtime.

As a proof-of-concept I'd like to add a simple assertion on the output of `Table.count` for one of the database tables, and then build more complicated cases from there.

I'm running into the following error/exception -- and have enabled `indexeddbshim`'s DEBUG option to provide a bit more context:

```bash
$ make tests
...
<snip>
...
Trying to fetch data for Index SELECT "value" FROM "S__sync^Nodes" WHERE "I_is^Master" NOT NULL AND "I_is^Master" = ? LIMIT 1 [ 'È-40110000000000' ]
TransactionInactiveError: A request was placed against a transaction which is currently not active, or which is finished. 
Trace: 
    at logError (/home/james/Documents/reciperadar/frontend/node_modules/indexeddbshim/src/DOMException.js:213:34)
    at createDOMException (/home/james/Documents/reciperadar/frontend/node_modules/indexeddbshim/src/DOMException.js:300:9)
    at Function.__assertActive (/home/james/Documents/reciperadar/frontend/node_modules/indexeddbshim/src/IDBTransaction.js:535:15)
    at IDBObjectStore.add (/home/james/Documents/reciperadar/frontend/node_modules/indexeddbshim/src/IDBObjectStore.js:506:20)
    at eval (webpack://%5Bname%5D/./node_modules/dexie/dist/dexie.mjs?:2952:44)
    at new Promise (<anonymous>)
    at Object.mutate (webpack://%5Bname%5D/./node_modules/dexie/dist/dexie.mjs?:2904:20)
    at Object.mutate (webpack://%5Bname%5D/./node_modules/dexie/dist/dexie.mjs?:3895:38)
    at eval (webpack://%5Bname%5D/./node_modules/dexie/dist/dexie.mjs?:1480:31)
    at checkTableInTransaction (webpack://%5Bname%5D/./node_modules/dexie/dist/dexie.mjs?:1342:20)
...
<snip>
...
  1 failing

  1) test
       load database state from document:
     OpenFailedError: TransactionInactiveError A request was placed against a transaction which is currently not active, or which is finished
 TransactionInactiveError: A request was placed against a transaction which is currently not active, or which is finished
```

I've had a look at https://github.com/dfahlander/Dexie.js/issues/712 and https://github.com/dfahlander/Dexie.js/issues/701#issuecomment-392756378 -- the latter mentions that `Native async await can result in that a transaction is committed too early (which is also the case in Firefox. This is due to incompatiblity between IndexedDB(Shim) and the native Promise). Transpiled async await still fine.` -- and I wonder if that could be relevant to the problem here.

I'd appreciate any assistance on how to investigate this further and any possible workarounds or fixes!  I've been attempting some very basic debugging by editing the local `node_modules` directory copy of `dexie` and am glad to provide any other assistance figuring out the cause.

jayaddison:
The cause of #1126 seems to relate to conflict between the (re)use of transactions on the `_syncNodes` Dexie-internal table after the `ready` trigger fires but before the database has completed opening.

The attempted fix here replaces two separate read & then write operations with a single-pass `modify` operation over the `db._syncNodes` internal table -- relying on [`Table.orderBy`](https://dexie.org/docs/Table/Table.orderBy()) to ensure that existing master sync node state is assessed before the 'local'  sync node record is updated -- within a  [`Dexie.ignoreTransaction`](https://dexie.org/docs/Dexie/Dexie.ignoreTransaction()) -- to retain atomicity, while avoiding re-use of the 'current' transaction.

Resolves #1126

### DIFF
--- a/addons/Dexie.Observable/src/Dexie.Observable.js
+++ b/addons/Dexie.Observable/src/Dexie.Observable.js
@@ -243,7 +243,7 @@ function Observable(db) {
             }
             // Add new sync node or if this is a reopening of the database after a close() call, update it.
             return db.transaction('rw', '_syncNodes', () => {
-                var currentMaster = db._syncNodes
+                return db._syncNodes
                     .where('isMaster').equals(1)
                     .first(currentMaster => {
                         if (!currentMaster) {
@@ -256,23 +256,18 @@ function Observable(db) {
                             currentMaster.isMaster = 0;
                             return db._syncNodes.put(currentMaster);
                         }
-                    });
-                Promise.all([currentMaster]);
-
-                // Add our node to DB and start subscribing to events
-                var currentNode = db._syncNodes.add(mySyncNode.node).then(function() {
-                    Observable.on('latestRevisionIncremented', onLatestRevisionIncremented); // Wakeup when a new revision is available.
-                    Observable.on('beforeunload', onBeforeUnload);
-                    Observable.on('suicideNurseCall', onSuicide);
-                    Observable.on('intercomm', onIntercomm);
-                });
-                Promise.all([currentNode]);
-
-                return Dexie.ignoreTransaction(function() {
-                    // Start polling for changes and do cleanups:
-                    pollHandle = setTimeout(poll, LOCAL_POLL);
-                    // Start heartbeat
-                    heartbeatHandle = setTimeout(heartbeat, HEARTBEAT_INTERVAL);
+                    }).then(()=>{
+                        // Add our node to DB and start subscribing to events
+                        return db._syncNodes.add(mySyncNode.node).then(function() {
+                            Observable.on('latestRevisionIncremented', onLatestRevisionIncremented); // Wakeup when a new revision is available.
+                            Observable.on('beforeunload', onBeforeUnload);
+                            Observable.on('suicideNurseCall', onSuicide);
+                            Observable.on('intercomm', onIntercomm);
+                            // Start polling for changes and do cleanups:
+                            pollHandle = setTimeout(poll, LOCAL_POLL);
+                            // Start heartbeat
+                            heartbeatHandle = setTimeout(heartbeat, HEARTBEAT_INTERVAL);
+                        });
                 });
             }).then(function () {
                 cleanup();
--- a/addons/Dexie.Observable/src/Dexie.Observable.js
+++ b/addons/Dexie.Observable/src/Dexie.Observable.js
@@ -243,17 +243,20 @@ function Observable(db) {
             }
             // Add new sync node or if this is a reopening of the database after a close() call, update it.
             return db._syncNodes.put(mySyncNode.node).then(Dexie.ignoreTransaction(() => {
+                // By default, this node will become master unless we discover an existing, up-to-date master
                 var mySyncNodeShouldBecomeMaster = 1;
                 return db._syncNodes.orderBy('isMaster').reverse().modify(existingNode => {
                     if (existingNode.isMaster) {
-                        // Master have been inactive for too long
-                        // Take over mastership
                         if (existingNode.lastHeartBeat < Date.now() - NODE_TIMEOUT) {
+                            // Existing master record is out-of-date; demote it
                             existingNode.isMaster = 0;
                         } else {
+                            // An existing up-to-date master record exists, so it will remain master
                             mySyncNodeShouldBecomeMaster = 0;
                         }
                     }
+                    // Assign the local node state
+                    // This is guaranteed to apply *after* any existing master records have been inspected, due to the orderBy clause
                     if (existingNode.id === mySyncNode.node.id) {
                         existingNode.isMaster = mySyncNodeShouldBecomeMaster;
                     }
--- a/addons/Dexie.Observable/src/Dexie.Observable.js
+++ b/addons/Dexie.Observable/src/Dexie.Observable.js
@@ -255,6 +255,12 @@ function Observable(db) {
                             mySyncNodeShouldBecomeMaster = 0;
                         }
                     }
+
+                    // TODO: Why is this safety check required?
+                    // Without it, this test fails: https://github.com/dfahlander/Dexie.js/blob/77f0b08c58784bfaaf9e2f5e26dd4e1d7b7d3094/test/tests-exception-handling.js#L411-L494
+                    // ... due to an unhandled global TypeError: Cannot read property 'id' of null
+                    if (!mySyncNode.node) return;
+
                     // Assign the local node state
                     // This is guaranteed to apply *after* any existing master records have been inspected, due to the orderBy clause
                     if (existingNode.id === mySyncNode.node.id) {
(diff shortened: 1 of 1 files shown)


jayaddison:
NB: This does pass existing tests (locally here, at least), but additional test coverage would be worthwhile to prove the fix and prevent regressions (assuming it's a reasonable/correct approach, which it may not be!)

addons/Dexie.Observable/src/Dexie.Observable.js
```diff
@@ -242,7 +242,7 @@ function Observable(db) {
                 });
             }
             // Add new sync node or if this is a reopening of the database after a close() call, update it.
-            return db.transaction('rw', '_syncNodes', () => {
+            return Dexie.ignoreTransaction(() => {
```
dfahlander: Could we keep the transaction here to make the addition of syncNode atomic but just add Dexie.ignoreTransaction() around it?

```js
return Dexie.ignoreTransaction(() => db.transaction('rw', '_syncNodes', () => {
    ...
}));

Comment (jayaddison) [reply addons/Dexie.Observable/src/Dexie.Observable.js]: Alas not - it seems like some of the logic within the existing `transaction` block still causes the transaction to become inactive, even when it's wrapped (rather than replaced) by `ignoreTransaction`.

Is it possible that the [`db._syncNodes.where` query](https://github.com/dfahlander/Dexie.js/tree/77f0b08c58784bfaaf9e2f5e26dd4e1d7b7d3094/addons/Dexie.Observable/src/Dexie.Observable.js#L240-L242) closes the transaction early, for some reason?

From debugging a bit more, it seems that any operations on `db._syncTables` (either read or write) within the `rw` transaction and after the `where` query triggers the `TransactionInactiveError`.

Comment (jayaddison) [reply addons/Dexie.Observable/src/Dexie.Observable.js]: NB: the requirement to keep the atomic behaviour makes sense; I'm trying a few other approaches at the moment to see whether it's possible to retain the transaction.

File: addons/Dexie.Observable/src/Dexie.Observable.js [hunk 2]
```diff
@@ -256,18 +256,23 @@ function Observable(db) {
...
-                            Observable.on('beforeunload', onBeforeUnload);
-                            Observable.on('suicideNurseCall', onSuicide);
-                            Observable.on('intercomm', onIntercomm);
-                            // Start polling for changes and do cleanups:
-                            pollHandle = setTimeout(poll, LOCAL_POLL);
-                            // Start heartbeat
-                            heartbeatHandle = setTimeout(heartbeat, HEARTBEAT_INTERVAL);
-                        });
+                    });
+                Promise.all([currentMaster]);
```
jayaddison: The intent behind this serialization was to attempt to keep the parent transaction 'active', as I have a sense that it is being closed as soon as the read operation (`where`) completes, and that as a result the `db._syncNodes` write operations fail later on.

I'm not super familiar with JavaScript promises so I'm not completely confident that this is still performing the correct set of operations in the correct order..

jayaddison: (this approach was indeed incorrect, as discovered by the `dexie-observable` unit test coverage)

addons/Dexie.Observable/src/Dexie.Observable.js
```diff
@@ -254,19 +254,23 @@ function Observable(db) {
...
-                        return db._syncNodes.add(mySyncNode.node).then(function() {
-                            Observable.on('latestRevisionIncremented', onLatestRevisionIncremented); // Wakeup when a new revision is available.
-                            Observable.on('beforeunload', onBeforeUnload);
-                            Observable.on('suicideNurseCall', onSuicide);
-                            Observable.on('intercomm', onIntercomm);
-                            // Start polling for changes and do cleanups:
-                            pollHandle = setTimeout(poll, LOCAL_POLL);
-                            // Start heartbeat
-                            heartbeatHandle = setTimeout(heartbeat, HEARTBEAT_INTERVAL);
+                        db.transaction('rw!', '_syncNodes', () => {
```
jayaddison: This 'works', but needs a bit of scrutiny as to whether it retains atomicity:

- https://dexie.org/docs/Dexie/Dexie.transaction()#specify-reusage-of-parent-transaction
- https://dexie.org/docs/Dexie/Dexie.transaction()#implementation-details-of-nested-transactions

Ideally some test coverage to check & prove atomicity would be nice.

jayaddison: Another related thought: this is currently essentially performing a get-and-set within a transaction to ensure atomicity.

Would it be possible instead to use a single write operation ([`Table.modify`](https://dexie.org/docs/Collection/Collection.modify()) or similar) to perform the required updates?

That way there would be no read-only query required within the transaction, and the transaction would not risk becoming inactive as a result of that query completing before any writes occur.

Our requirements are to:

- Set the current node's `isMaster` flag when no record with `isMaster` and an up-to-date heartbeat timestamp is present
- Unset the `isMaster` flag on a record if the heartbeat timestamp is out-of-date
- Write the current node record into the table

The challenge is that the value we write for the current node record depends on the contents of a potentially existing (or potentially nonexistent) record with `isMaster` set and potentially out-of-date heartbeat timestamp.

jayaddison: It does look like the `isMaster` field is an [indexable type](https://dexie.org/docs/Indexable-Type) (ref: [definition as `Number`](https://github.com/dfahlander/Dexie.js/blob/77f0b08c58784bfaaf9e2f5e26dd4e1d7b7d3094/addons/Dexie.Observable/src/Dexie.Observable.d.ts#L96)), so perhaps it would be possible to use [`db._syncNodes.orderBy(...).modify(...)`](https://dexie.org/docs/Table/Table.orderBy()) to retrieve existing syncnodes in order, maintain an external variable to store the 'current node should become master' state, and then use a function argument to [modify](https://dexie.org/docs/Collection/Collection.modify()) to write the expected object record(s).

jayaddison:
@dfahlander Apologies for any notification noise from this and the associated issue over the past few days; it's taken a while to work out what's been going on, and attempt different approaches to a solution :)

I think this is ready for another look now; I'm becoming more confident in the correctness of this after understanding the purpose of the code a little better, and given the test coverage in Dexie itself and also when using it alongside the [dexie-1126-repro](https://github.com/jayaddison/dexie-1126-repro) example case.

dfahlander (approved):
Thanks for the deep dive in this issue and for finding a solution.

jayaddison:
No problem, thanks!

PR #1165, opened 2020-11-11
Built-in observability: Collection.subscribe()

dfahlander:
Make Collection & Table comply to es-observable's subscribe method.

https://github.com/tc39/proposal-observable

Should be compatible with RxJS. Needs to be verified.

Idea is to enable live queries without having to go with dexie-observable, since that addon affects performance by adding the mutations to a _changes table - which we really do not need here.

The limitation of this built-in observable capability is currently that it does not react ot DB mutations on other tabs/windows. However, is this really desired??? Do we want inactive tabs to start working and take performance from the computer when things happen in DB? If really needed, that functionality could be easily patched on top of this library by firing the new db.on.mutate event when storage event happens. Or in the case of a worker mutating data, it could be patched by letting the worker post message to frontend code that would fire the db.on.mutate event.

### DIFF
--- a/src/classes/collection/collection.ts
+++ b/src/classes/collection/collection.ts
@@ -12,6 +12,7 @@ import { hangsOnDeleteLargeKeyRange } from "../../globals/constants";
 import { ThenShortcut } from "../../public/types/then-shortcut";
 import { Transaction } from '../transaction';
 import { DBCoreCursor, DBCoreTransaction, DBCoreRangeType, DBCoreMutateResponse, DBCoreKeyRange } from '../../public/types/dbcore';
+import { Observer, Subscription } from '../../public/types/observable';
 
 /** class Collection
  * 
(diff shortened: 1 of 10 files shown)


rakeshpai:
I think we can look at other tabs / workers in a second pass, but if we have that it'd be really awesome. Doesn't need to happen automatically if you want to keep that out of this lib - I think it'll be ok to have the worker write some glue code to postMessage to the main thread, or import an additional package to make that happen. Windows that users open don't have references to each other, so postMessaging across windows isn't always possible,  but that can be worked around using `window.on('storage')` or other such hacks.

So, all in all, this is awesome! We can look at cross-tab / worker messaging later.

dfahlander:
Thanks a lot for the good feedback :) The implementation of extending observation to changes made in other tabs or in workers could just fire the new 'mutate' event manually when it has the knowledge of foreign mutations.

For example, a solution that only works between tabs but now between workers would be able to do something like this:
```js
db.on('mutate', mutatedParts => {
  Object.keys(mutatedParts).forEach(table => {
    localStorage.setItem(`idb-observer:${db.name}:${table}`, Math.random().toString());
  });
});
window.addEventListener('storage', ({key}) => {
  db.on.mutate.fire({[key]: true});
});
```
A similar solution could be done for workers.

rakeshpai:
This looks really cool!

WebWorkers are always 'owned' by some page, and serviceWorkers own all pages, so a similar mechanism would work everywhere using `postMessage` of some sort. Doesn't need to be part of core - it's just a few lines of glue that can be put up on a gist or packaged into a separate lib.

One little nitpick: The type definition of `MutatedParts` only seems to report which table has changed, not which record has changed or what the new state of the record is. I guess this might not be relevant when you're looking at this as an Observable (I only have superficial familiarity with observables). But when you're looking at this as the building blocks for a react hook, an additional optimisation is that when a record has changed and you know the new state, the UI can be re-rendered without having to query the DB. So, if `MutatedParts` could contain not just the table, but the record and its new state, that would be great. dexie-observable does this already, I think. I haven't fully understood this implementation to know if that's possible here, so I could be completely misunderstanding this.

As both you and I mentioned in the original thread, re-querying the DB may not be a big deal, but it'd be great if we can skip it anyway. Afaict, this avoiding-requerying optimisation can only be used if an existing record has mutated, not if a new record is added or a record is deleted, so this might not be super useful, but it'll be nice to not re-render if we can avoid it.

I could be totally misunderstanding this PR - please correct me if I'm wrong. Regardless, I'm excited about this! Thanks a ton for moving this forward!

dfahlander:
Agree - it's hard to do something in the library for the worker use case - so better to require the glue and document examples of how to do it. I'll probably add more info in the MutatedParts type.

It is possible to compute the result of an existing query in memory, based on the mutations made - but - it would be a bit big and messy. But I think at least it would be nice to be able to optimize key-based queries (queries on the primary key) by recording the keys in the value of MutatedParts instead of a boolean. Index based queries could also be recomputed if we would record the entire mutations, but I would be worried it could starve memory to do that all the time - especially for large bulk mutations.

dfahlander:
This PR will be superseeded by #1172 that makes a broader approach than just letting a Collection or Table be an observable and instead let any JS expression as `db.friends.get(x)` be observable. It can even observe the result of any function that returns a promise that originate from Dexie queries.

Issue #1163, opened 2020-11-03
Export loadUntilWeGotEnoughData or JsonStream


fishbone1:
Can you please export `loadUntilWeGotEnoughData()` and/or `JsonStream()` in your npm module?

Background:
We delete our database before importing the file. But before importing, we need to know if the backup file is valid.
Your method `loadUntilWeGotEnoughData()` uses JsonStream to fetch the relevant JSON part which allows you to check if the file is valid, but the deletion must occur between this validation and the import itself.

We want to go even further and implement a "dry run" which imports the data into a temporary db. If this is successfull we can delete the actual db and start the actual import.

To achieve this, we wrote a method which is almost identical to `importDB`. The only difference is, that it uses the temporary db name:

```ts
export async function dryRunImportDB(exportedData: Blob | JsonStream<DexieExportJsonStructure>, options?: StaticImportOptions): Promise<Dexie> {
  options = options || {}; // All booleans defaults to false.
  options.acceptNameDiff = true; // DRY RUN
  const CHUNK_SIZE = options!.chunkSizeBytes || (DEFAULT_KILOBYTES_PER_CHUNK * 1024);
  const stream = await loadUntilWeGotEnoughData(exportedData, CHUNK_SIZE); // CANNOT IMPORT THIS METHOD
  const dbExport = stream.result.data!;
  const db = new Dexie(dbExport.databaseName + '_dryrun'); // DRY RUN
  db.version(dbExport.databaseVersion).stores(extractDbSchema(dbExport));
  await importInto(db, stream, options);
  return db;
}
```
But this doesn't work, because we cannot `import` `loadUntilWeGotEnoughData()` in line 5.

The temporary dry-run db doesn't exist before running the import. The key thing here is this line:
```ts
  db.version(dbExport.databaseVersion).stores(extractDbSchema(dbExport));
```
It initializes the object stores. This is only possible, because `loadUntilWeGotEnoughData()` has the necessary information.

Exporting `extractDbSchema()` would be helpful as well, but it's very simple and can be implemeted on our own.

We tries to implement `loadUntilWeGotEnoughData()` but it creates a `JsonStream` which is also not exported. Only its type information is exported. Exporting `JsonStream` would be very useful, because it allows to write your own backup file validation logic.

fishbone1:
For anyone who is interested, here is the workaround for my Angular-project, but it should work with other typescript projects as well:

[...]

PR #1166, opened 2020-11-11
dexie-export-import: Peek import file (Solution to #1163?)

dfahlander:


addons/dexie-export-import/src/import.ts
```diff
@@ -45,6 +45,12 @@ export async function importDB(exportedData: Blob | JsonStream<DexieExportJsonSt
   return db;
 }
 
+export async function peakImportFile(exportedData: Blob): Promise<Partial<DexieExportJsonStructure>> {
+  const CHUNK_SIZE = (DEFAULT_KILOBYTES_PER_CHUNK * 1024);
+  const stream = await loadUntilWeGotEnoughData(exportedData, CHUNK_SIZE);
```
fishbone1: `loadUntilWeGotEnoughData` imho does too much here and doesn't allow to handle file problems/errors on your own. It has two responsibilities: Read the header AND validate the file. My simple suggestion is to take the upper part of `loadUntilWeGotEnoughData`, i.e. the part before validation.

So I'd replace this line by:

```
  const stream = ('slice' in exportedData ?
    JsonStream<DexieExportJsonStructure>(exportedData) :
    exportedData);

  while (!stream.eof()) {
    await stream.pullAsync(CHUNK_SIZE);

    if (stream.result.data && stream.result.data!.data)
      break;
  }
```

fishbone1 (changes requested):

dfahlander:
I added a new commit that follows your advice and also:

* Reads headers in chunks of 5 kb instead of 1 MB.
* Returns an object where the data.data array is not present (as it would be half-baked).

fishbone1:
Yes deleting data.data makes absolutley sense.

Advantage of your whole solution is that you don't need to expose `JsonStream`. I think it's a good thing that it's not part of the public API.

PR #1177, opened 2020-11-23
Vue.js example

andyhaskell:
This PR adds a Vue.js example to the Dexie.js `samples` folder. With Vue 3 now out, I decided to make it in Vue 3 with vue-cli. It is largely based on the component layout in the React example to have those examples be similar.

Excellent library by the way. I find it really straightforward and satisfying to use!


dfahlander (approved):
Nice work! Thanks!

PR #1196, opened 2020-12-15
Prohibit master "surprises" by not just running PRs on firefox

dfahlander:


dfahlander:
Closing this for now. Might reopen later. This would put a bit more traffic onto browserstack and we might see more randomly failing tests for shaky IE11 on PRs.

PR #1197, opened 2020-12-16
Never treat array properties as delta-objects in update hook

### LINKED ISSUE #1195: Update with array as value adds number objects
I might be completely missing something but this is odd behavior to me:
```
const field = 'authors'

  await db.books.update(id, {
     [field]: [{ foo: 'bar' }, { foo2: 'bar2'} ]
   })
```

This ends up with `authors.0` and `authors.1` but it should just be an array.
What's going on here?

dfahlander:
When diffing 2 objects, treat array properties as intrinsic objects and never go into them to treat array indices as if they were object props.

Resolves #1195.

### DIFF
--- a/src/functions/utils.ts
+++ b/src/functions/utils.ts
@@ -259,7 +259,7 @@ export const getValueOf = (val:any, type: string) =>
                 const bpTypeName = toStringTag(bp);
 
                 if (apTypeName === bpTypeName) {
-                    if (intrinsicTypeNameSet[apTypeName]) {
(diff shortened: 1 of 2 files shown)


aiingstan:
Hi, could you advise when will this be released? I'm using version `3.0.3`, it still stores array modifications as `xx.1` format. Thanks!

dfahlander:
It's released in latest alpha version. Do `npm install dexie@next` to get it.

PR #1212, opened 2021-01-20
Expose close event

jsrilapan:
Add the close event to the DbEvents list, so close event can be subscribed by doing this,
db.on('close'). This close event only fires if the database is unexpectedly closed.


dfahlander:
Thanks a lot!

PR #1217, opened 2021-01-30
Update package homepage link proto (HTTP to HTTPS)

jayaddison:
Extremely minor; reduces HTTP redirects during sessions that open the link.  A lot of NPM packages seem to link to their homepages via HTTPS (this is visible when using some of the NPM CLI tools), so this stood out and looks like a small change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
   "engines": {
     "node": ">=6.0"
   },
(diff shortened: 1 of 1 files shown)


dfahlander:
Thanks! Missed it when adding https support.

jayaddison:
No problem, thanks!

PR #1242, opened 2021-03-04
Issue1240

### LINKED ISSUE #1240: useLiveQuery with multi-entry index
I think this is a bug, but it may be user error. useLiveQuery doesn't seem to update the query when I `add` a new item to the collection.

As a repro, I [forked the live query sandbox](https://codesandbox.io/s/hidden-cherry-9gmir) and changed the `age` to a multi-entry index (basically, I made age store a history of the birthdays instead of the current age).

If you load that sandbox and set the max age to 100, then add a person with an age of 21, the person will not show up in the friends list until you edit the max age to reboot the query.

If this isn't a bug, is there another way I can tell the live query to update when I add a value?

dfahlander:
Reproduce and fix #1240 - support for multiEntry indices in liveQuery.

### DIFF
--- a/src/live-query/observability-middleware.ts
+++ b/src/live-query/observability-middleware.ts
@@ -267,13 +267,19 @@ function trackAffectedIndexes(
     function extractKey(obj: any) {
       return obj != null ? ix.extractKey(obj) : null;
     }
+    const addKeyOrKeys = (key: any) => ix.multiEntry && isArray(key)
+      // multiEntry and the old property was an array - add each array entry to the rangeSet:
+      ? key.forEach(key => rangeSet.addKey(key))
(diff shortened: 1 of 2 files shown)


dfahlander:
The solution should work the same as specified for multiEntry index in the [IndexedDB specification 2.0](https://www.w3.org/TR/IndexedDB-2/#object-store-storage-operation) where it states in bullet no 5 and 6:

5: _If index’s multiEntry flag is unset, or if index key is not an array key then store a record in index containing index key as its key and key as its value...._

6: _If index’s multiEntry flag is set and index key is an array key, then for each subkey of the subkeys of index key store a record in index containing subkey as its key and key as its value..._

PR #1244, opened 2021-03-05
Break out liveQuery()

dfahlander:
My try to break out liveQuery() from dexie into its own library and let it emit Rx observables.

Vision with this PR:
1. Publish a new dexie@3.1.0-beta.x and base live-query package on that version of dexie.
2. Publish the new library and test it
3. Remove Observable, Subscription and liveQuery from dexie along with the test-live-query.
4. Make dexie-react-hooks dependent on the new library and import { liveQUery } from it instead of dexie.

Why:
* Better to base our observables on existing library (rxjs) than letting Dexie come with an Observable type.
* Don't want to make dexie dependent on rxjs itself. But ok to make a new library dependent on it.


dfahlander:
Dexie 3.2.0 will keep having liveQuery() built-in. Closing this PR.

Issue #854, opened 2019-05-28
dexie-export-import: JSON formatting is invalid in exports


travddm:
Dexie.js version 2.0.4
dexie-export-import version 1.0.0-alpha.10
-> Using all default options

While exporting and importing a large database (>600 MB), the JSON generated by the export was invalid.

After parsing the file with jsonlint, the problem seemed to be that there were missing commas between objects in large arrays. (expected between objects: "},{" vs found between objects: "}{") Throughout the whole file, I only found about 30 instances of these missing commas.

After replacing each "}{" with "},{", the database imported properly without any problems.

I suspect this issue may only be apparent in larger sets of data, and I also suspect it has something to do with objects being separated right on the start/end of each streaming chunk.

dfahlander:
Thanks. I'll try to find a way to trigger this in a unit test and fix. Feel free to help me with the unit test if you'd have the time to make a PR

dfahlander:
I found the issue and it should be fixed now. Try with the latest release on npm (1.0.0-beta.11)

PR #1271, opened 2021-04-13
Point to the correct start and end indices

zhenyanghua:
This fixes the chunk being sliced to its own copy.

### DIFF
--- a/addons/dexie-export-import/src/tson.ts
+++ b/addons/dexie-export-import/src/tson.ts
@@ -78,5 +78,8 @@ TSON.finalize = async (items?: any[]) => {
     }
   }
   // Free up memory
-  blobsToAwait = [];
+  // Todo - this is the problematic one that causes the blobs to be reset everytime the `finalize` is
+  //  called, where the chunks slicing are happening, which causes the data slicing with the wrong indices.
+  //  need to figure out a different way to GC.
+  // blobsToAwait = [];
 }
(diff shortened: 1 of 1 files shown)


dfahlander:
Thanks @zhenyanghua for the PR!

I see the test did fail for this fix. The reason might be an error in the test itself.

To understand what issue is being fixed, please file an issue that describes it. Also, it would be great if you could add another promisedTest() in basic-tests.ts or edge-cases.ts that reproduces the issue with the existing code. You could do that in this ongoing PR and refer to it from your issue.

zhenyanghua:
The edge tests failed on CI (even locally all tests pass) due to:
>   ✖ chunkedExport (issue #854)
      Firefox 85.0 (Linux x86_64)
    Got error: NotReadableError: File could not be read
    NotReadableError: File could not be read

From MDN, it says: 
> The file or directory cannot be read, typically due to permission problems that occur after a reference to a file has been acquired (for example, the file or directory is concurrently locked by another application).

I suspect it may be this operation that puts the file on lock:

https://github.com/dfahlander/Dexie.js/blob/b6d621ef25075f4f54cfdd923834bf939351cd93/addons/dexie-export-import/src/tson.ts#L60

@dfahlander I am puzzled now with this issue. Do you have any thoughts on the issue?

dfahlander:
Closing. PR too old.

PR #1272, opened 2021-04-13
Improved handling of sub-modifications and dotted keys

### LINKED ISSUE #1270: Modification object in updating hook not correct when changing array
See https://jsfiddle.net/ygaprkbe/2/

In the console you will see `{}` but it should be the object in the `update` call.

So far I've only seen this when updating items inside of an array.

Also, most likely a related bug: https://jsfiddle.net/ygaprkbe/3/

The modifications object here is `{ foo.bar: 444 }` which seems similar to this old bug #1195

Christilut:
This fixes #1270

I did the following:
- Updated `getValueOf` to use JSON.stringify with a custom replacer so it handles ArrayBuffer properly
- Changed rv[] to `setKeyByPath` so the dotted notation should not occur anymore
- Using REPLACEMENT_CHAR (arbitrary value, might need to be changed?) instead of a dot so a dot in keys can still be used (test #966 covers that)
- Added tests including one for arraybuffer in sub-objects to confirm the stringify replacer works

Please let me know if something is not right

### DIFF
--- a/test/tests-crud-hooks.js
+++ b/test/tests-crud-hooks.js
@@ -593,7 +593,7 @@ spawnedTest("updating using Table.put()", function*(){
         op: "update",
         key: 1,
         obj: {id:1, address: {city: 'A'}},
-        mods: {"address.city": "B"},
+        mods: {address: { city: "B" }},
         updatedObj: {id:1, address: {city: 'B'}},
     }], ()=>db.transaction('rw', db.tables, function* (){
         db.table1.put({id:1, address: {city: 'A'}}); // create
@@ -616,7 +616,7 @@ spawnedTest("updating using Table.bulkPut()", function*(){
         op: "update",
         key: 1,
         obj: {id:1, address: {city: 'A'}},
-        mods: {"address.city": "B"},
+        mods: {address: {city: "B"}},
         updatedObj: {id:1, address: {city: 'B'}},
     }], ()=>db.transaction('rw', db.tables, function* (){
         db.table1.put({id:1, address: {city: 'A'}}); // create
@@ -639,7 +639,7 @@ spawnedTest("updating using Table.update()", function*(){
         op: "update",
         key: 1,
         obj: {id:1, address: {city: 'A'}},
-        mods: {"address.city": "B"},
+        mods: {address: {city: "B" }},
         updatedObj: {id:1, address: {city: 'B'}},
     }], ()=>db.transaction('rw', db.tables, function* (){
         yield db.table1.add({id:1, address: {city: 'A'}}); // create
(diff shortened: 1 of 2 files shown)


test/tests-crud-hooks.js
```diff
@@ -616,7 +616,7 @@ spawnedTest("updating using Table.bulkPut()", function*(){
         op: "update",
         key: 1,
         obj: {id:1, address: {city: 'A'}},
-        mods: {"address.city": "B"},
+        mods: {address: {city: "B"}},
```
dfahlander, line 619: What is the reason `{"address.city": "B"}` is changed to `{address: {city: "B"}}`? In the original code, we are just replacing the "city" property of the inner object, and that must also be possible to do. The changed mods means we are replacing the whole "address" property to be exactly {city: "B"} which is different.

It is important that the existing unit tests (with `{"address.city": "B"}`) are still working so it would be better to leave existing unit tests as they are.

Christilut:
I still want to finish this but I do need some feedback

dfahlander:
I understand this can be an issue but I think I have given some questions to you that I haven't got an answer on.

Specifically:

What is the reason `{"address.city": "B"}` is changed to `{address: {city: "B"}}`? In the original code, we are just replacing the "city" property of the inner object, and that must also be possible to do. The changed mods means we are replacing the whole "address" property to be exactly `{city: "B"}` which is different.

It is important that the existing unit tests (with `{"address.city": "B"}`) are still working so it would be better to leave existing unit tests as they are.

dfahlander:
Also: Why do we need to have a different replacement char?

Christilut:
I did answer in the conversation above but I'll try to explain more:

I understand that it's important to keep tests working but in this case I think the test is wrong because if the input is `{id:1, address: {city: 'A'}}` then so should the modifications. With this input, the modifications being `{"address.city": "A"}` is incorrect and this is the whole issue.
If this means that the sub-property gets replaced entirely then we need to change it to a merge/assign. I can do that and write a test for it.

The replacement character is needed because if the replacement character is a dot then `address.city` will be interpreted as `address: { city: .. }`. In this case I chose `~~` but this was just a suggestion and we should probably change it to something that will most likely never be used in a property although I'm confident that `~~` is good.

dfahlander:
Thanks for your work on this. I was only worried for making a too complex solution. Also, it would change the current behaviour of diffing nested object properties if we'd always specify that the entire outermost property was changed while it only was a sub property that was actually changed. Not very important but I see no reason to change that behaviour - especially as it could affect production code for lots of people.

However, the root issue #1270 is very important to fix at is an unacceptable behaviour. I ended up submitting an alternate PR (#1332) where I started out by copying the unit test from this PR. I modified the test slightly to comply with current dexie behaviour of nested prop diffs on objects.

PR #1283, opened 2021-04-21
Preserve CryptoKey objects when adding to auto-incrementing table

### LINKED ISSUE #1280: Cannot add CryptoKeys to table with auto-incrementing primary key
## Context & Repro

Here's a minimal repro:

```js
const db = new Dexie("MyDatabase");
db.version(1).stores({
    keys: "++id",
});

// Create a CryptoKey
const someCryptoKey = (
  await window.crypto.subtle.generateKey(
    {
      name: "RSA-OAEP",
      modulusLength: 4096,
      publicExponent: new Uint8Array([1, 0, 1]),
      hash: "SHA-256",
    },
    true,
    ["encrypt", "decrypt"]
  )
).privateKey;
const objToAdd = { key: someCryptKey };

console.log(objToAdd.key instanceof CryptoKey); // true


// Attempt to add it to the table
db.keys.add(objToAdd);

const storedObj = await db.keys.get(1);
console.log(storedObj.key instanceof CryptoKey); // unexpectedly false
```

The above only repro's when using an auto-incrementing primary key and when the keyPath is not provided in `objToAdd`.

It appears that the root cause is in these lines:
https://github.com/dfahlander/Dexie.js/blob/d6fe43eeaf45fa5992c2288a0bd40d7cd05fef17/src/classes/table/table.ts#L283-L288

Which uses a deep-clone: 
https://github.com/dfahlander/Dexie.js/blob/d6fe43eeaf45fa5992c2288a0bd40d7cd05fef17/src/functions/workaround-undefined-primkey.ts#L6

And it looks like the deep-clone algorithm doesn't work with `CryptoKey` objects.

## Potential approaches for fix?

1. Does `workaroundForUndefinedPrimKey()` need to perform a deep clone? 

2. If we *must* perform a deep clone, would adding `CryptoKey` to the list of "intrinsic types" make the cloning algorithm use the original `CryptoKey` object?
https://github.com/dfahlander/Dexie.js/blob/d6fe43eeaf45fa5992c2288a0bd40d7cd05fef17/src/functions/utils.ts#L193-L194

seshrs:
As discussed in #1280, this PR:

- Adds `CryptoKey` to the list of IntrisicTypes to be preserved by the deep cloning algorithm.
- Modifies the "undefined primary key" workaround for auto-incrementing tables to only run on instances of user-defined classes (as opposed to plain-old JavaScript objects (POJO)).
- Adds unit tests for the above.

Suggestions and feedback are welcome. Thanks @dfahlander for your quick response and suggestions 😃 

Closes #1280.

### DIFF
--- a/src/functions/utils.ts
+++ b/src/functions/utils.ts
@@ -191,7 +191,7 @@ export function flatten<T> (a: (T | T[])[]) : T[] {
 
 //https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
 const intrinsicTypeNames =
-    "Boolean,String,Date,RegExp,Blob,File,FileList,ArrayBuffer,DataView,Uint8ClampedArray,ImageData,Map,Set"
+    "Boolean,String,Date,RegExp,Blob,File,FileList,ArrayBuffer,DataView,Uint8ClampedArray,ImageData,Map,Set,CryptoKey"
     .split(',').concat(
         flatten([8,16,32,64].map(num=>["Int","Uint","Float"].map(t=>t+num+"Array")))
     ).filter(t=>_global[t]);
--- a/src/functions/workaround-undefined-primkey.ts
+++ b/src/functions/workaround-undefined-primkey.ts
@@ -1,7 +1,18 @@
 import { deepClone, delByKeyPath, getByKeyPath } from './utils';
 
+// This workaround is needed since obj could be a custom-class instance with an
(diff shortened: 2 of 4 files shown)


dfahlander:
Thanks! I will go through before releasing dexie@3.1.0-alpha.10.

dfahlander:
The CI error seems to be some unrelated failure. I'll trigger a new build.

dfahlander:
I did some updates to this PR and created a new PR #1292 with your + my changes.

PR #1287, opened 2021-04-22
Fix Version signatures

### LINKED ISSUE #1286: Incorrect type signature for Version.upgrade
As a public interface, `Version` has the following signature for `upgrade`:
https://github.com/dfahlander/Dexie.js/blob/master/src/public/types/version.d.ts#L5
```
upgrade(fn: (trans: Transaction) => void): Version;
```

As a class and in the documentation, it has the following signature:
https://github.com/dfahlander/Dexie.js/blob/master/src/classes/version/version.ts#L63
```
upgrade(upgradeFunction: (trans: Transaction) => PromiseLike<any> | void): Version
```

The difference in the argument type prevents from returning the promise when using the upgrade function and can lead to bad usage.

genezys:
Fixes #1286

### DIFF
--- a/src/classes/version/version.ts
+++ b/src/classes/version/version.ts
@@ -8,7 +8,7 @@ import { exceptions } from '../../errors';
 import { createTableSchema } from '../../helpers/table-schema';
 
 /** class Version
- * 
+ *
  * http://dexie.org/docs/Version/Version
  */
 export class Version implements IVersion {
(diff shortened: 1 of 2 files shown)


dfahlander:
Thanks! Looks good.

genezys:
I also changed the implementation for `stores` to be consistent with the signature and documentation.

PR #1314, opened 2021-05-26
Check for global before checking its prop

gotjoshua:
Trying to get dexie syncable round trip in https://blog.stackblitz.com/posts/introducing-webcontainers/

### DIFF
--- a/addons/Dexie.Observable/src/Dexie.Observable.js
+++ b/addons/Dexie.Observable/src/Dexie.Observable.js
@@ -519,7 +519,7 @@ try {
 //
 // Map window events to static events in Dexie.Observable:
 //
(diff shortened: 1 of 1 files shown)


dfahlander:
Thanks.

gotjoshua:
> Thanks.

welcome! thanks for your amazing product/project!

when do you expect this change to be published to npm (with the next version i guess)?

dfahlander:
Will do a new release within the coming 2 weeks. The publish script will then also publish dexie-observable if we have bumped its version number. I do that now so it won't be missed.

dfahlander:
It's released on in dexie-observable@3.0.0-beta.11.

Issue #1236, opened 2021-02-22
FileHandle (from File System Access API) is not serialized unless primary key is not inbound


jgentes:
This is a reproduction of this working scenario using `idb-keyval`: https://glitch.com/edit/#!/filehandle-indexeddb?path=script.js%3A1%3A0

With Dexie, using the following code:

```
db.version(1).stores({
  test: '++id'
});

const [fileHandle] = await window.showOpenFilePicker();
db.test.add({ fileHandle })
```

This results in a record with `{ fileHandle: {} }` (empty object)

However, if I do not name the primary key, it serializes the object properly:

```
db.version(1).stores({
  test: '++'
});

const [fileHandle] = await window.showOpenFilePicker();
db.test.add({ fileHandle })
```

Result: `{ fileHandle: FileSystemFileHandle... }`

![image](https://user-images.githubusercontent.com/1149470/108771276-46a3d280-7510-11eb-810e-b96e6b4a47a2.png)

Fortunately this workaround is ok for me, but seems like it may be a bug, and certainly something that may trip people up.

jgentes:
Looks like this also works:

```
db.version(1).stores({
  test: '&name'
});
```

dfahlander:
Seems to me as a browser bug. idb-keyval does not name the primary key either so they probably don't hit the bug.

datdinhquoc:
> Seems to me as a browser bug. idb-keyval does not name the primary key either (no [inbound](https://dexie.org/docs/inbound) key) so they probably don't hit the bug.

is it possible to solve this bug? i would like to save to indexeddb file handle as value in a doc, eg:
```
{
    name: "some file name",
    handle: FileHandleInstanceHere
}
```

The only way currently is to use idb-keyval, when using Dexie and set the file handle as value, the file handle turns into blank object `{}`.

dfahlander:
You can use [outbound keys](https://dexie.org/docs/inbound#example-of-outbound-primary-key) also with Dexie.

nonovd:
No problems on **add operations** with either an outbound key or a key with an unique index.

However, on **update operations**, the file handle will be lost.

To continue with the previous example:

```javascript
async updateName () {
  return await db.fileHandles.update(2, { name: 'newName' })
}
```
_name_ will be updated and _handle_ will be an empty object.

| #   |      Key     |  Value |
|:----------|:-------------:|------:|
[...]

PR #1327, opened 2021-06-03
Support for FileSystemFileHandle when cloning

dfahlander:
Add FileSystemFileHandle to the list of intrinsic type names - so that object cloning works for FileSystemFilehandles.


dfahlander:
Resolves #1236.

PR #1329, opened 2021-06-05
Transaction.abort does not rollback changes

mushishi78:
I've written a test to demonstrate. The documentation [here](https://dexie.org/docs/Transaction/Transaction.abort()) states that abort `// Will discard all changes`, but it doesn't seem to for me. I ran this is chrome.


dfahlander:
Thanks! I'll look into it soon!

dfahlander:
We should have had this test long time before.

mushishi78:
Thanks for looking into it so fast 😅

dfahlander:
It was important to fix, and a perfect repro. I tried your repro on previous version and the bug has been there for ages. I'm a bit surprised no one has found it earlier but maybe people have just thrown an error instead to work around it.

Issue #1333, opened 2021-06-13
uniqueKeys produces duplicate values with virtual index


davidnx:
### Problem description

`Collection.uniqueKeys` incorrectly produces duplicate results when operating on a virtual index. This is indicative that not only is it producing incorrect results, but it also is not running efficiently as it is considering more items than would be necessary to find the right answer.

It's worth mentioning that Doxie docs mention virtual indexes should support the entire feature set of Dexie. Per [Compound Indexes](https://dexie.org/docs/Compound-Index):

> Virtual indexes have the full support for any operation, just like ordinary indexes, such as equalsIgnoreCase(), anyOf(), etc.

### Example

```ts
db.version(1).stores({
    metrics: "id,[name+time]"
});

await db.metrics.add({ id: "id1", name: "a", time: 1 });
await db.metrics.add({ id: "id2", name: "b", time: 2 });
await db.metrics.add({ id: "id3", name: "a", time: 3 });
const result = await db.metrics.orderBy("name").uniqueKeys();
// Produces result = ["a", "a", "b"]; Expected ["a", "b"]
```

davidnx:
I added a unit test that confirms this issue on latest master. Also repros on latest released versions `3.0.3` and `3.1.0-beta.13`

See: https://github.com/dfahlander/Dexie.js/commit/017cdea34eb188b3e0a166f2aca33a56bc56f75c

davidnx:
Thanks @dfahlander!

PR #1334, opened 2021-06-14
Added test case for #1333 - uniqueKeys produces duplicate values with…

dfahlander:
… virtual index


dfahlander:
Completed in #1345

Issue #1342, opened 2021-06-27
Bug: dexie-export-import


vikrend:
When importing if `chunkSizeBytes` is small enough not to fit all table rows and flag `clearTablesBeforeImport` is set to true, import clears table before each chunk put. This results in only last chunk being imported.

As i suppose the expected behavior is table being cleared once - before first put.

[Sandbox demo](https://codesandbox.io/s/quizzical-wing-0bs3n?file=/src/index.js)

PR #1343, opened 2021-06-29
Bugfix: fix importInto clearTablesBeforeImport option (issue #1342)

vikrend:
Bugfix + unit test.
In this commit import clears all db tables if flag is provided. Or is it intended to clear only those tables which are being imported?


dfahlander:
Very nice repro and fix! Thanks!

Issue #1018, opened 2020-04-13
When will dexie support relaxed durability transactions


kasnihuang:
Chrome Exposes an optional relaxedDurability parameter on IDBDatabase.transaction to control flushing to disk. And will available at 83.
How long will Deixe support this feature after the 83 version is released?
https://chromestatus.com/feature/5730701489995776

dfahlander:
Firefox uses relaxed durability in all IDB transactions without opting in to it, while Chrome and other browsers uses strict durability. Now Chrome wants to enable relaxed durability as an opt-in. From a developers point of view, it would be nice if the behavior could be same along implementations and that it would have the same default value.

For now, Dexie does not specify the `{relaxedDurability: true}` (as it seems to be named when looking at [the C++ changes in chromium](https://chromium-review.googlesource.com/c/chromium/src/+/1775557)) option when we create a transaction. But if we would support it, I would suggest that transactions by default would have relaxedDurability turned on, and that it could be possible to force strict durability by explicitly turning it off, and that this would work the same across browsers. One idea would be to introduce the "d" letter in the Dexie-style transaction mode. A transaction could be marked "rwd", which would specify {relaxedDurability: false} as IDBTransactionOptions, while unless "d" is among the transaction mode, we'd use {relaxedDurability: true} by default. This would probably be ignored in browsers where the transaction options are not supported.

So for Dexie, this could be a two-step feature:

Step 1: Start specifying {relaxedDurability: true} on all transactions. Will only be used by chrome 83 and later. All others would ignore it. Firefox would use relaxed durability anyway.

Step 2: Allow explicit durable transactions, by introducting the "d" letter in dexie's transaction mode. This would only be respected by chrome 83 and later, and maybe also in a later version of Firefox if they want to go for it. Safari would keep using durable transactions in all cases unless they also want to use the flag.

alexburlton-sonocent:
This is now possible per https://github.com/dfahlander/Dexie.js/pull/1367, and has been released in `3.2.0-beta.3`.

frankleng:
@dfahlander how does relaxed durability affecting live query updates?
I was seeing delay in a somewhat large dataset fed by `liveQuery`

dfahlander:
In dexie 4, transactionless updates affect live queries immediately in the default case - the update itself is applied to ongoing queries directly in memory. In this case, we have invented an even more relaxed durability. However, this only works with simple queries such as below, between, equals, etc but also startsWith.

[...]

PR #1367, opened 2021-07-23
Add ability to specify durability for transactions in Chrome

alexburlton-sonocent:
Implementing step 1 as described in [this issue](https://github.com/dfahlander/Dexie.js/issues/1018)

This PR updates dexie so that we can pass a `chromeTransactionDurability` when constructing a `Dexie` instance to drill down into any transactions that are used when interacting with it. This is a special flag that was added by Chrome per [this feature](https://chromestatus.com/feature/5730701489995776) (source code for the change is [here](https://chromium.googlesource.com/chromium/src/+/d762124e2b4090a7985cddc5438678de7900fcc4/third_party/blink/renderer/modules/indexeddb/idb_transaction_options.idl)). For other browsers (or versions of Chrome older than 83) it will have no impact.

I've added tests to check that the raw `IDBTransaction` has the desired state based on what is passed, and have also used the `npm link` strategy to verify that there's a performance boost to our app when using the modified Dexie code. We wrote a simple test that inserted blobs into a table in a loop, under three scenarios:

1) Using `Dexie.Table`
2) Using raw `IDBTransaction` with `{ durability: 'relaxed' }`
3) Using raw `IDBTransaction` with `{ durability: 'default' }`
 
Before the change, the timings for 1) and 3) aligned, with 2) being about twice as fast. Linking the changes in this PR, the timings for 1) reduced in-line with 2). Happy to share the source code for the test we ran if it's of any interest.


dfahlander:
Thanks! I couldn't get the same performance benefits when I tested this out earlier. Probably depends a lot about the use case. Not sure that changing the default is ok in a minor version so for 3.x we could allow the option in Dexie constructor but keep the default durability and plan to change the default for next major (4.x).

alexburlton-sonocent:
> Thanks! I couldn't get the same performance benefits when I tested this out earlier. Probably depends a lot about the use case. Not sure that changing the default is ok in a minor version so for 3.x we could allow the option in Dexie constructor but keep the default durability and plan to change the default for next major (4.x).

Yeah that makes sense - I guess I was hoping to get away with the lazy option :stuck_out_tongue: 

I'll look at plumbing the durability properly so it's configurable. Also, looks like tests are failing because they get run on multiple browsers (so my new tests fails on Firefox for instance - see [here](https://app.travis-ci.com/github/dfahlander/Dexie.js/jobs/526412105)). Is there an established pattern for having tests run only on certain browsers that I can follow to get this passing?

alexburlton-sonocent:
Updated to plumb as a `DexieOption` - also renamed to `chromeTransactionDurability` so it's super clear it's a Chrome-specific thing.

I also extracted a separate test file for this stuff since I need to construct a Dexie db per test with different options. Still got the outstanding question of how to skip these tests for non-chrome so we can get the Travis build to pass, otherwise I think this is ready for review/release in a minor 3.x version :crossed_fingers:

dfahlander:
Great! There is a supports() function in [dexie-unittest-itils.js](https://github.com/dfahlander/Dexie.js/blob/master/test/dexie-unittest-utils.js) that you could extend. It's also ok to do user-agent sniffing here. If the browser doesn't support the feature, let the test succeed with ok(true, description).

alexburlton-sonocent:
Ah nice! I've put in an `isChrome` var which seems to work for me. It detects newer versions of Edge, too - but that's fine since it's basically the same browser these days (it's just reskinned chrome, so the tests will pass there I imagine).

alexburlton-sonocent:
Is there anything else you need from me on this? Would be awesome to get it merged and released if possible so we can see the impact it has on our users (if this doesn't provide a sufficient enough performance boost, the sooner we find out the sooner we can do more investigation / look into other avenues) :crossed_fingers:

dfahlander:
Currently it's just the comment I made about the variable isChrome not working (which basically makes it not trying your tests on chrome either).

dfahlander:
I thought I commented with a jsfiddle link but can't find it now. Anyway, I've tried the isChrome variable in a fiddle and it evaluated to false on all browsers.

alexburlton-sonocent:
Ah interesting - yeah, for me locally in chrome it was skipping the tests too it turns out. It's odd - I tested the flag using chrome dev tools and it worked in there :shrug: 

I've simplified the flag and it seems to be working now :+1:

dfahlander:
Thanks! It's merged.

alexburlton-sonocent:
> Thanks! It's merged.

Fantastic! Thanks for your help and super speedy responses - this has been an excellent contribution experience from my perspective :grin: 

Any idea when a new release might be published? Excited to get this out into the wild so we can see how it affects our metrics in production :drooling_face:

dfahlander:
Happy to hear that 👍 
I will start the release-script this afternoon. It does some additional tests on complementary browsers but unless any script fails, it would be out soon.

dfahlander:
Sorry, the release-script failed on a regression test on Chrome 79 (listed in our config file as "chrome_oldest_supported"). Obviously, it's a chrome but probably without the support for this option. Either, we could update the chrome_oldest_supported to a version that has support for this option, or we could extend the test condition to also verify chrome version (could be extracted using a regexp on navigator.userAgent). Another option would be to allow undefined as a valid value of trans.idbtrans.durability (as it would only be set to "default" or "relaxed" if it's a chrome that supports it)

dfahlander:
Is it ok if I fix it by allowing trans.idbtrans.durability to be undefined in all three tests?

alexburlton-sonocent:
> Is it ok if I fix it by allowing trans.idbtrans.durability to be undefined in all three tests?

Yep, sounds good!

dfahlander:
It's released. If you'd get inspiration to update the docs for Dexie constructor (https://dexie.org/docs/Dexie/Dexie) that would be a nice thing. If not, that's ok also and we try make sure fixing it later. The current doc is stone-age old with clumsy HTML table. It's ok to update it to a markdown table if you have that inspiration.

alexburlton-sonocent:
> It's released. If you'd get inspiration to update the docs for Dexie constructor (https://dexie.org/docs/Dexie/Dexie) that would be a nice thing. If not, that's ok also and we try make sure fixing it later. The current doc is stone-age old with clumsy HTML table. It's ok to update it to a markdown table if you have that inspiration. (Use the "Edit on github" button to edit the page and it will generate a PR)

Thanks so much! Updating some docs feels like the least I can do - I'll aim to get to it at some point this afternoon :smile:

PR #1379, opened 2021-08-15
fix(bulk-delete): correctly define keys type

### LINKED ISSUE #1364: Cannot use primaryKeys to pass into bulkDelete in Angular 11
I am writing a generic function in Angular 11 to delete records based on a where clause. However, I am unable to pass the results of a .where(where).primaryKeys() to a bulkDelete:

image

image

Based on the [docs](https://dexie.org/docs/Table/Table.bulkDelete()), it seems like I should be able to use the primary keys array as the parameter for the bulk delete?

Currently working around it by setting keys to any[] which seems to work in Chrome.

diegocasmo:
Closes #1364


dfahlander:
Thanks for this PR!

Issue #613, opened 2017-11-13
Chrome - Random Closing Database issue (hard/impossible to reproduce)


jokamax:
Hi,

Some users on Chrome (version > 51 maybe before) encouters in rare case : Database closing (invalidStateError).

The issue (chrome issue) seem to be known : [Connection closed on Chrome (IndexedDB) in long sessions](https://github.com/localForage/localForage/issues/581)

Is there a clean way on dexie to catch that and re-open (without losing data) ?

Or is it possible to make an evolution to automaticaly retry connexion on not wanted closing (option in open init, and the retry is invisible for the code/user during actions on the database) ?

Thanks by advance,

Jonathan

dfahlander:
It's possible to listen to errors globally and take actions. Try listening globally on the ['unhandledrejection'](http://dexie.org/docs/Promise/unhandledrejection-event) event. Log the error name to console. When you have the error name, you may add code that recovers from it by reopening db. If you find a solution for this chrome issue this way, please share it!

dfahlander:
Need to say as well, that you will also need to log the event when catching it explicitly, not just rely on the unhandledrejection event, as it will not fire for catched promises.

trackedsupport:
> This is happening for us too. It seems like some data in indexeddb is lost.
> 
> > Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.

I'm getting this error a lot more now that I switch from saving blobs to saving ArrayBuffers. Im not sure how to fix though. I never close the DB and there is no new version of the DB.

mikekreeki:
For anyone else bumping into this, I took a workaround implemented here https://github.com/yjs/y-indexeddb/blob/ef0a100a40e0102fa4f1ec32130fd75348d4efd2/src/y-indexeddb.js
and seems to work fairly well.

```js
let DB;

const initDB = () => {
  const db = new Dexie(DATABASE_NAME);

  db.version(1).stores(mapping);

  return db;
};

const getDB = () => {
  if (!DB) {
    DB = initDB();
  }

  const idb = DB.backendDB();

  if (idb) {
    try {
      // Check if if connection to idb did not close in the meantime
      idb.transaction('tasks').abort();
    } catch (e) {
      DB.close();
      DB = initDB();
    }
  }

  return DB;
}

// Use it
getDB().table('tasks').add(record)
```

dfahlander:
[...]

PR #1398, opened 2021-09-29
A workaround for issue 613

dfahlander:


trackedsupport:
Is there a way I can start using this today?

dfahlander:
> Is there a way I can start using this today?

Clone or fork this repo and checkout branch **workaround-issue-613** . Then do npm install and npm run build in your clone. Then depending if you use npm or yarn

yarn:

1. `yarn link` ( when you are in your dexie clone)
2. cd to your app that depends on dexie and do `yarn link dexie`

npm:

1. `npm link` ( when you are in your dexie clone)
2. cd to your app that depends on dexie and do ,,`npm link dexie`

dfahlander:
Please post back here or on issue #613 on how the fix behaves in production.

trackedsupport:
Thanks I got that to work but I dont see how this will deploy. In fact, I dont even see how its linked to my dexie clone with any files pointing to the local dexie clone.

dfahlander:
This is a generic npm/yarn procedure applicable for any project. Please refer to the docs of npm link for a detailed explanation. The PR will also be released in a beta once 3.2.0 stable is out.

mikekreeki:
Just for the context, I found this in `localforage` codebase, seem like the same workaround at the first glance https://github.com/localForage/localForage/blob/master/src/drivers/indexeddb.js#L393-L395

dfahlander:
A new release dexie@3.2.1-beta.1 was published just now with this PR included. Please try it and report back whether it's a valid workaround for #613 or not.

trackedsupport:
Testing now!

trackedsupport:
I have not seen this issue since I deployed the fix/workaround. Thank you so much!

trackedsupport:
i did get this one time. Still much better than what it was! My application does stay on one page for a long time. 

Screen Shot 2021-11-24 at 9 01 54 PM

trackedsupport:
So I pulled in a local version of Dexie to play around with the `PR1398_maxLoop`. I set it to 3000 and I am still getting the error. Now they are all from iOS Safari or Chrome so that tells me it could be a bug with iOS. Unfortunately, I don't know when it will be fixed and I don't know what I can do about it except tell the user to stop their recording and refresh the page which is a poor experience.

dfahlander:
I've reviewed but as I can see it, that line can only be hit after failed opening of the database. As soon as the workaround is executed, it calls db.open() which synchronically sets db._state.openComplete to false. I can see no race condition because whatever operation in Dexie will from that point start waiting for the open operation to complete. The only case I can see is that for some reason the database has failed to reopen - which would be another situation than what we've worked around in this one - in contrast to the case when exception is thrown when creating a transaction, we don't know whether that situation would possible to work around by retrying or not. It would be interesting to know what the value of db._state.dbOpenError.name is when this line is hit! Notice that this line can also be hit in case a subscriber to db.on('ready') have failed so we cannot just retry every time it happens, but in case we can distinguish the reason by inspecting what error did occur it might be worth it to do a retry for db.open() at this point somehow (in your clone/fork to see whether it would improve the statistics)

dfahlander:
@trackedsupport Did I understand you correctly that the error only occurs on iOS now. As iOS Chrome is just a rebranded Safari this would mean that the chrome-specific bug would be worked around here but it could be a Safari issue we're hitting instead. Is it possible to get the user agent or safari version of the failing clients?

trackedsupport:
Hey, So it happens 99% of the time with iOS Safari and Chrome. I am able to recreate with my iPad. 

`db._state.dbOpenError.name` = "UnknownError"

trackedsupport:
So I've been running a bunch of tests today with my iPad that kept reproducing the error. I added a `db.open()` right before line 1269 in the image above but that did not help. 

I decided to upgrade my iPad to 15.4 from 15.3 since the issue didn't occur on my iPhone anymore. Sure enough, that worked. So I'm thinking they fixed it in the latest iOS release!

dfahlander:
👍 Thanks for sharing! That's a good sign we're doing what's possible here for Chrome and that the Safari-specific issue might be solved in Apple webkit.

Issue #1392, opened 2021-09-19
Are large indexes bad? (and what would be a sane limit?)


Sjoerd82:
I have some fairly wide tables (many attributes), which I use in a variety of (regular/auto-complete) drop down boxes, thus requiring only a few attributes from the whole table (limit/offset is not always an option).

Initially I thought about creating smaller tables with only the required data, but when I thought about it, there probably isn't a real difference in creating and retrieving the same subset of data from an index on the large table.

At this point I'm creating indexes consisting of about 3-6 attributes (40-80 bytes), and fetch only the key using `db.myTable.orderBy('[id+type_id+group_id+name+type_name]')`.

So far performance seems to be good, ±30-40% better vs. toArray() (in a limited test)

I wonder what the impact is on the table size, and if there are any possible negative side-effects on this approach.
So the question is:
- Is it bad to have wide indexes, and what would a sane limit to the number of attributes in an index / or size? Is an index with 10 attributes too long? (I come from a relational db world, no-sql is new to me)
- Can I assume that the size of an index on a table is simply the size of the data in that index? And this goes on top of the already existing data (ie. data is duplicated? or pointers?)
- Are these actually Dexie questions or IndexedDb questions?

For context, the worst-case scenario I can imagine would be a 50,000 rows table with an index of 10 attributes / 200 bytes.

dfahlander:
Sorry for late reply. I wish I could answer these questions but I can only contribute with what I would believe based on my experiences with IndexedDB. If you need hard numbers, it will need to be tested.

IndexedDB works very similar to an SQL database. Especially regarding indexes which are BTree based indexes. The amount of storage used will vary across implementations and fragmentations might make it take up more space than the exact index size just as it would in any other BTree database. I know some other databases (postgres/SQLServer etc) have some limit per index key (2712 bytes in postgres, it seems, 1700 bytes in SQLServer and 3072 for MySQL). IndexedDB has no limit (but I think it should because performance suffer when using 10MB long index keys as some people do). Some people index image data or even movies they store in ArrayBuffers (which might be several megabytes long) (for no reason).

In your case though, it's about 10 small and sane indexes. I would think that as long as your total data that is included in the index, is with margin below the limits of mysql/postgres/SQLServer, they would be ok. But if you want the hard numbers, do some test and measure it using [StorageManager.estimate()](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/estimate). In case you do this, feel free to share the numbers in this issue!

[...]

PR #1412, opened 2021-10-14
Switch to JS implementation of indexedDB.cmp().

dfahlander:
This one is about 30 times faster than indexedDB.cmp()
+ it removes the global dependency of indexedDB for RangeSet,
used by liveQuery().

Chrome:

| Implementation | Number of iterations | Time taken (string) | Time taken (number) | Time taken (Uint8Array) |
|---|---|---|---|---|
| indexedDB.cmp() | 1_000_000 | 550 ms | 503 ms | 790 ms |
| plain JS version | 1_000_000 | 17.5 ms | 13 ms | 27 ms |
| *factor faster* | | *31* | *39* | *29* |

Firefox:

| Implementation | Number of iterations | Time taken (string) |
|---|---|---|
| indexedDB.cmp() | 1_000_000 | 370 ms |
| plain JS version | 1_000_000 | 50 ms |

Safari:

| Implementation | Number of iterations | Time taken (string) |
|---|---|---|
| indexedDB.cmp() | 1_000_000 | 110 ms |
| plain JS version | 1_000_000 | 22 ms |

Bundle sizes

| Dexie version | Minified | Minified & gzipped |
|---|---|---|
| dexie@3.0.3 | 71.13 kb | 22.1 kb |
| dexie@3.2.0-rc.2 (with liveQuery()) | 78.3 kb | 25.1 kb |
| dexie@3.2.0-... with also JS cmp | 79 kb | 25.3 kb |

### DIFF
--- a/src/dbcore/keyrange.ts
+++ b/src/dbcore/keyrange.ts
@@ -1,12 +1,10 @@
-import { domDeps } from '../classes/dexie/dexie-dom-dependencies';
-import { getMaxKey } from '../functions/quirks';
 import { DBCoreKeyRange, DBCoreRangeType } from '../public/types/dbcore';
 
 export const AnyRange: DBCoreKeyRange = {
   type: DBCoreRangeType.Any,
   lower: -Infinity,
   lowerOpen: false,
-  get upper() { return getMaxKey(domDeps.IDBKeyRange) },
+  upper: [[]],
   upperOpen: false
 }
 
--- a/test/tests-misc.js
+++ b/test/tests-misc.js
@@ -1,6 +1,6 @@
 import Dexie from 'dexie';
 import {module, stop, start, asyncTest, equal, deepEqual, ok} from 'QUnit';
-import {resetDatabase, spawnedTest, promisedTest, supports} from './dexie-unittest-utils';
+import {resetDatabase, spawnedTest, promisedTest, supports, isIE, isEdge} from './dexie-unittest-utils';
 
 const async = Dexie.async;
 
@@ -321,6 +321,11 @@ asyncTest ("#1079 mapToClass", function(){
 });
 
 asyncTest("PR #1108", async ()=>{
+    if (isIE || isEdge) {
+        ok(true, "Disabling this test for IE and legacy Edge");
+        start();
+        return;
+    }
     const origConsoleWarn = console.warn;
     const warnings = [];
     console.warn = function(msg){warnings.push(msg); return origConsoleWarn.apply(this, arguments)};


src/dbcore/dbcore-indexeddb.ts
```diff
@@ -39,12 +39,9 @@ export function getKeyPathAlias(keyPath: null | string | string[]) {
 
 export function createDBCore (
   db: IDBDatabase,
-  indexedDB: IDBFactory,
   IdbKeyRange: typeof IDBKeyRange,
   tmpTrans: IDBTransaction) : DBCore
 {
-  const cmp = indexedDB.cmp.bind(indexedDB);
```
diegocr, line 46: Hello, i was very excited when i first read about those performance improvements, however when taking a look at the code i got some doubts.

- Have to tried to perform those benchmarks without the use of `Function.bind` (?)
- This JS version of cmp seems only faster for small amounts of data, and specially when the data isn't equal from the first few bytes (?)

Check this out:

```js
b1=cryptHelper.getRandomValues(473)
b2=cryptHelper.getRandomValues(473)

a1 = () => indexedDB.cmp(b1,b2)
a2 = () => compareUint8Arrays(b1,b2)
```

```js
> console.time('dd');for(let i = 1e6; i--; a1());console.timeEnd('dd');
< dd: 799.961181640625 ms

> console.time('dd');for(let i = 1e6; i--; a2());console.timeEnd('dd');
< dd: 5.804931640625 ms
```

All good here, with 473bytes of data and not equal.

```js
b2=b1
> console.time('dd');for(let i = 1e6; i--; a1());console.timeEnd('dd');
< dd: 822.533203125 ms

> console.time('dd');for(let i = 1e6; i--; a2());console.timeEnd('dd');
< dd: 702.003173828125 ms
```

Still reasonable here when the data is equal, but...

```js
b1=cryptHelper.getRandomValues(4730)
b2=b1

> console.time('dd');for(let i = 1e6; i--; a1());console.timeEnd('dd');
< dd: 1101.98681640625 ms

> console.time('dd');for(let i = 1e6; i--; a2());console.timeEnd('dd');
< dd: 8124.588134765625 ms
```

Thoughts? Not sure if i did overlooked something, those tests were made under Chrome 96 on Windows.

src/functions/cmp.ts
```diff
@@ -1,18 +1,89 @@
...
+
+export function compareUint8Arrays(
+  a: Uint8Array,
+  b: Uint8Array
+) {
+  const al = a.length;
+  const bl = b.length;
+  const l = al < bl ? al : bl;
+  for (let i = 0; i < l; ++i) {
+    if (a[i] !== b[i]) return a[i] < b[i] ? -1 : 1;
```
diegocr, line 55: ...as a follow up to my previous comment, here's an experiment with `DataView`

```js
function compareUint8Arrays(a,b) {
    const dv1 = new DataView(a.buffer);
    const dv2 = new DataView(b.buffer);
    let idx = 0;
    let bytes = Math.min(dv1.byteLength, dv2.byteLength);
    if (bytes > 4) {
        let len = bytes >> 2;
        while (idx < len) {
            let b1 = dv1.getUint32(idx);
            let b2 = dv2.getUint32(idx);
            if (b1 !== b2) return b1 < b2 ? -1 : 1;
            ++idx;
        }
        bytes -= idx << 2;
    }
    if (bytes > 2) {
        let len = bytes >> 1;
        while (idx < len) {
            let b1 = dv1.getUint16(idx);
            let b2 = dv2.getUint16(idx);
            if (b1 !== b2) return b1 < b2 ? -1 : 1;
            ++idx;
        }
        bytes -= idx << 1;
    }
    while (idx < bytes) {
        let b1 = dv1.getUint8(idx);
        let b2 = dv2.getUint8(idx);
        if (b1 !== b2) return b1 < b2 ? -1 : 1;
        ++idx;
    }
    return dv1.byteLength === dv2.byteLength ? 0 : dv1.byteLength < dv2.byteLength ? -1 : 1;
}
```

```js
b1=cryptHelper.getRandomValues(4730)
b2=b1

> console.time('dd');for(let i = 1e6; i--; a2());console.timeEnd('dd');
< dd: 1382.230224609375 ms
```

... 83% faster with 4KB of data (?) but still slower than `indexedDB.cmp`

dfahlander: If I get it right, the conclusion of all this is that:

* JS based comparison still beats native one by far for numbers and strings (and probably Dates and reasonably short arrays)
* The only situation where we can see native indexedDB.cmp beat the JS based one is when comparing large binaries that have an identical start sequence in their first 500 bytes.
* Binary keys still perform much better with the JS based algorithm than the native one as long as they are reasonably short or have reasonable randomness.

Now, when we are comparing keys, we only do it for indexed keys and properties and these shall not be large anyway (see [#1392](https://github.com/dexie/Dexie.js/issues/1392#issuecomment-927841921)) and there is almost never any good reason to index large binaries. A typical binary index represent either a primary key of maybe 16 (random) bytes, or digests that may be between 8-32 bytes and have the same randomness. There might be other use cases, like indexing paths of tree structures that may be longer, and having identical start sequences, but probably never that large. Indexing binaries larger than say 1500 bytes would create other performance issues anyway because it bloats the BTree in the database - which is the reason that most databases limit the permitted size of BTree indexes to 1700-3500 bytes depending on DB engine.

dfahlander:
See my review comments - real world indexed binary keys are not large and therefore the plain JS version outperforms indexedDB.cmp() by far in most cases anyway. There might be use cases for indexed binary keys up to around 1000 bytes at most but indexing larger binary keys than that is bad practice anyway (See https://github.com/dexie/Dexie.js/issues/1392#issuecomment-927841921). Therefore I don't think it's worth it to optimize the code for the cost of increased bundle size and possible bugs.

diegocr:
Thanks David, that's very helpful information. Also, i was apparently under the wrong impression this comparator was used for more stuff than just keys, so i do fully understand this now and i do agree about using the current JS version as-is.

fwiw, i just took a look at https://github.com/dexie/Dexie.js/search?q=cmp and found the following which you may want to move as well, as they're tied to the `Where` clauses we may get a little performance boost there as well (?)

https://github.com/dexie/Dexie.js/blob/fe682ef24568278c3b31d9d6c93de095d4b77ae8/src/classes/where-clause/where-clause-constructor.ts#L33-L35

https://github.com/dexie/Dexie.js/blob/e865e80a3c05821db40ecc3c64157104d345b11e/src/classes/table/table.ts#L130

Cheers!

dfahlander:
Thanks! Good you saw that. I'll replace them.

PR #1428, opened 2021-11-08
Should use Record as type instead Object

Dumeng:
Refer to the typescript-eslint rule [ban-types](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/ban-types.md#default-options). Use `Record<string, unknown>` as type instead of `Object` or `object`.


dfahlander:
Thanks for the PR! It will however conflict with #1424 where this is also adressed. There I'm using template literal types so that object keys needs to be on the model or in a deep structure of the object (for example, passing  `{"address.city": "London"}` is ok if the the model has `{address: { city: string }}`).

PR #1459, opened 2021-12-03
Abortable live query

dfahlander:
Aborts liveQueries if unsubscribing from a liveQuery observable while querier is being executed.

Currently, when a live query is unsubscribed from while the querier is being executed, we let the querier continue it's execution but ignore it's result.

This PR will abort the querier eagerly for the purpose of unloading the db from unnecessary requests.

There might be a risk that aborting ongoing transactions could trigger possible bugs in native indexedDB implementations. Another risk is that there would be unhandled promise rejections when unsubscribing. These things needs to be tested before merging this query.


dfahlander:
Closing this PR as it is a part of #1718 instead and will be merged in that.

PR #1465, opened 2021-12-22
Prevent exception on broadcast channel failure

Chocobozzz:
When cookies are disabled, `new BrodacastChannel()` throws an exception

### DIFF
--- a/src/live-query/enable-broadcast.ts
+++ b/src/live-query/enable-broadcast.ts
@@ -5,25 +5,9 @@ import {
 } from '../globals/global-events';
 import { propagateLocally, propagatingLocally } from './propagate-locally';
 
-if (typeof BroadcastChannel !== 'undefined') {
-  const bc = new BroadcastChannel(STORAGE_MUTATED_DOM_EVENT_NAME);
+const broadcastChannelSuccess = createBroadcastChannel()
 
-  //
-  // Propagate local changes to remote tabs, windows and workers via BroadcastChannel
-  //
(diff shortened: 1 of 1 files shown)


dfahlander:
Question 1: The fallback to use localStorage when BroadcastChannel throws, will it actually work when cookies are disabled? I could assume that also localStorage would throw or at least not work in that scenario?

Question 2: I don't understand the if clause:

```ts
if (!broadcastChannelSuccess || typeof self !== 'undefined' && typeof navigator !== 'undefined') { ... }
```
Seems that we'd always go for the fallback solution as long as self isn't undefined. Seems as a logical error. The idea is to use the fallback when broadCastChannel fails. Here it's combined with an or statement that is true in both windows and workers making the fallback solution being run in all cases but node.

Chocobozzz:
> Question 1: The fallback to use localStorage when BroadcastChannel throws, will it actually work when cookies are disabled? I could assume that also localStorage would throw or at least not work in that scenario?

Yes I think it won't work

> The idea is to use the fallback when broadCastChannel fails

Then we should use an `&&`?

```js
if (!broadcastChannelSuccess && typeof self !== 'undefined' && typeof navigator !== 'undefined') { ... }

```

dfahlander:
I don't feel comfortable with this. If BroadcastChannel isn't working when cookies are disabled, changes are that neither IndexedDB or localstorage works either. Taking care of this scenario needs a new issue where the situation is described, the expected and actual behaviors should also be described as well as which browsers it has been tested with and how to disable cookies in those browsers (how to reproduce).

PR #1510, opened 2022-02-25
Middleware request ergonomics

gotjoshua:
- [x] add keys to put and add requests ([5bd4a8](https://github.com/gotjoshua/Dexie.js/commit/5bd4a893738fc2a0e2e37be35245e338db28a888))
- [x] export getEffectiveKeys

### DIFF
--- a/src/hooks/hooks-middleware.ts
+++ b/src/hooks/hooks-middleware.ts
@@ -37,9 +37,13 @@ export const hooksMiddleware: Middleware<DBCore>  = {
           const {deleting, creating, updating} = dxTrans.table(tableName).hook;
           switch (req.type) {
             case 'add':
+              // Ensure that the req has keys
+              req.keys = getEffectiveKeys(downTable.schema.primaryKey, req)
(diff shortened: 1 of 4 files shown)


src/hooks/hooks-middleware.ts
```diff
@@ -37,9 +37,13 @@ export const hooksMiddleware: Middleware<DBCore>  = {
           const {deleting, creating, updating} = dxTrans.table(tableName).hook;
           switch (req.type) {
             case 'add':
+              // Ensure that the req has keys
+              req.keys = getEffectiveKeys(downTable.schema.primaryKey, req);
```
dfahlander, line 41: This would slow down all requests even when there are no hooks registered.
Not all hooks would need to have the keys so even if there was an 'adding' hook registered, it could hurt performance for hooks that wouldn't want this.

dfahlander:
Thanks for this PR. To export getEffectiveKeys() in public Dexie API would be ok but I would avoid adding keys to put and add requests generically as it could hurt performance.

PR #1559, opened 2022-04-30
Support for FileSystemDirectoryHandle when cloning

bjoeni:
Same issue mentioned in #1236 exists for `FileSystemDirectoryHandle`s as well


dfahlander:
Thanks!

PR #1576, opened 2022-05-30
4.0.0-alpha.3 - Reproduce: Node.js process not exiting when Dexie.js is imported

pubkey:
When Dexie.js is imported in a JavaScript file and run via Node.js, just importing the dexie.js library will prevent the process from exiting.

This happens with node.js >= 18.
It is not a problem with node.js 16 and below.

I think there is some side-effect like a setTimeout or setInterval that prevents the process from ever being finished.

I have this problem when using `4.0.0-alpha.3` in a test suite that is run via Node.js for faster test runs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ before_install: if [[ `npm -v` != 7* ]]; then npm i -g npm@7; fi
 addons:
   firefox: "85.0"
 node_js:
-  - "12"
+  - "18"
 matrix:
   fast_finish: true
 services:
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
     "test": "npm run build && npm run test:typings && npm run test:unit",
     "test:unit": "karma start test/karma.conf.js --single-run",
     "test:typings": "tsc -p test/typings-test/",
-    "test:debug": "karma start test/karma.conf.js --log-level debug"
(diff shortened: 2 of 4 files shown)


pubkey:
Would you accept a PR to switch to github actions?

dfahlander:
Thanks for the PR. There's no setTimeout or setInterval going on in the repo in any root code so it's a mystery how it keeps the process alive by just requiring "dexie". I suppose debugging the paths taken in enable-broadcast.ts and propagate-locally.ts when run in node 18 could be something to debug, in case node 18 exposes APIs that used to be DOM specific,  like BroadcastChannel or addEventListener.

> Would you accept a PR to switch to github actions?

Go ahead if you like, but I can't answer right away whether I would merge it as it would needs some more investigation in regards to action minutes per month.

pubkey:
@dfahlander thank you for that hint.
Indeed, commenting out the BroadcastChannel usage solves my issue.

```ts
// const bc = new BroadcastChannel(STORAGE_MUTATED_DOM_EVENT_NAME);
```

I will work on a fix.

pubkey:
I am proposing a fix here: https://github.com/dexie/Dexie.js/pull/1577
I see no risk in merging the fix, but the test in #1576 should not be merched until travis supports node.js 18 or until we switch to a different CI provider.

PR #1577, opened 2022-05-30
FIX #1576 Node.js process not exiting when Dexie.js is imported

pubkey:
A bugfix that ensures that importing Dexie.js in Node.js does not prevent the process from finishing. This is mostly relevant when Dexie.js is used in Node.js for unit tests.

Described problem without the fix: https://github.com/dexie/Dexie.js/pull/1576

Docs for the used `unref()` method:
https://nodejs.org/api/worker_threads.html#broadcastchannelunref

### DIFF
--- a/src/live-query/enable-broadcast.ts
+++ b/src/live-query/enable-broadcast.ts
@@ -8,6 +8,18 @@ import { propagateLocally, propagatingLocally } from './propagate-locally';
 if (typeof BroadcastChannel !== 'undefined') {
   const bc = new BroadcastChannel(STORAGE_MUTATED_DOM_EVENT_NAME);
 
+  /**
+   * The Node.js BroadcastChannel will prevent the node process from exiting
+   * if the BroadcastChannel is not closed.
+   * Therefore we have to call unref() which allows the process to finish
+   * properly even when the BroadcastChannel is never closed.
+   * @link https://nodejs.org/api/worker_threads.html#broadcastchannelunref
+   * @link https://github.com/dexie/Dexie.js/pull/1576
+   */
+  if (typeof (bc as any).unref === 'function') {
+    (bc as any).unref();
+  }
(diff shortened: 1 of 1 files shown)


dfahlander:
Thanks! Very interesting.

loureirorg:
@dfahlander First of all, thanks for this fantastic library. 

Can this fix be applied to version 3?

I'm running Jest tests with Node.js, and it gets stuck when I import the Dexie class. 

Here's the log:
```bash
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  MESSAGEPORT

    > 1 | import Dexie, { Collection, Table } from "dexie";

      at ../node_modules/dexie/src/live-query/enable-broadcast.ts:9:14
      at ../node_modules/dexie/dist/dexie.js:15:85
      at Object.<anonymous> (../node_modules/dexie/dist/dexie.js:18:3)
      at Object.<anonymous> (simple.test.ts:1:1)
```

According to the error log, the issue is in `enable-broadcast.ts` on line 9, which is this very same issue fixed on version 4:
https://github.com/dexie/Dexie.js/blob/master-3/src/live-query/enable-broadcast.ts#L9

My solution is to run Jest with `--forceExit`, but it would be great if you could apply this fix to version 3.

dfahlander:
I published a dexie 3.2.3 with the fix.

```
npm i dexie@3.2.3
```

loureirorg:
It works! Thank you for the fast fix.

jayaddison:
I think that this has fixed a long-standing issue with some never-exiting unit tests using Dexie v3 that I hadn't been able to figure out.  Thank you, @pubkey @loureirorg @dfahlander!

Dmatt18 (commented):
Hi Dose any one Know how to Get Wrapper Offline

PR #1598, opened 2022-07-21
trim trailing spaces

jimmywarting:
Trim all whitespaces.


dfahlander:
I plan to run all code against prettier at some point, but that will be done in a controlled moment. Closing this one because it will continue to diverge and be obsolete when all code has ran through prettier.

PR #1624, opened 2022-10-01
feat(d.ts): type creating hook subscriber return value

hekystyle:
Hi.

Based on what I found writen in [docs](https://dexie.org/docs/Table/Table.hook('creating')) ...

>  // If returning any value other than undefined, the returned value will be used as primary key

... I feel like return type of `subscriber` param of `creating` table hook can be significantly improved.

### DIFF
--- a/src/public/types/table-hooks.d.ts
+++ b/src/public/types/table-hooks.d.ts
@@ -19,7 +19,7 @@ interface DeletingHookContext<T,Key> {
 }
 
 interface TableHooks<T=any,TKey=IndexableType> extends DexieEventSet {
-  (eventName: 'creating', subscriber: (this: CreatingHookContext<T,TKey>, primKey:TKey, obj:T, transaction:Transaction) => any): void;
+  (eventName: 'creating', subscriber: (this: CreatingHookContext<T,TKey>, primKey:TKey, obj:T, transaction:Transaction) => void | undefined | TKey): void;
   (eventName: 'reading', subscriber: (obj:T) => T | any): void;
   (eventName: 'updating', subscriber: (this: UpdatingHookContext<T,TKey>, modifications:Object, primKey:TKey, obj:T, transaction:Transaction) => any): void;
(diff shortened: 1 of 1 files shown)


dfahlander:
Thanks! That's correct.

lvivski:
Hi @hekystyle @dfahlander, this change seems to be introducing type mismatch with [`strictFunctionTypes`](https://www.typescriptlang.org/tsconfig#strictFunctionTypes) enabled in Typescript. It is enabled by default in `strict` mode.

Example
```ts
import Dexie from "dexie";

class DB extends Dexie {
  public readonly someTable!: Dexie.Table<any, string>;
}

const db = new DB("name");

db.transaction("rw", db.someTable, () => {});
```

```
(property) DB.someTable: Dexie.Table<any, string>
No overload matches this call.
  Overload 1 of 6, '(mode: TransactionMode, tables: readonly (string | Table<any, IndexableType, any>)[], scope: (trans: Transaction & { someTable: Table<any, string>; }) => void | PromiseLike<...>): PromiseExtended<...>', gave the following error.
    Argument of type 'Table<any, string>' is not assignable to parameter of type 'readonly (string | Table<any, IndexableType, any>)[]'.
      Type 'Table<any, string>' is missing the following properties from type 'readonly (string | Table<any, IndexableType, any>)[]': length, concat, join, slice, and 15 more.
  Overload 2 of 6, '(mode: TransactionMode, table: string | Table<any, IndexableType, any>, scope: (trans: Transaction & { someTable: Table<any, string>; }) => void | PromiseLike<...>): PromiseExtended<...>', gave the following error.
    Argument of type 'Table<any, string>' is not assignable to parameter of type 'string | Table<any, IndexableType, any>'.
      Type 'Table<any, string>' is not assignable to type 'Table<any, IndexableType, any>'.
        Types of property 'hook' are incompatible.
          Types of parameters 'subscriber' and 'subscriber' are incompatible.
            Type 'void | IndexableType | undefined' is not assignable to type 'string | void | undefined'.
              Type 'number' is not assignable to type 'string | void | undefined'.ts(2769)
```

PR #1631, opened 2022-10-20
Fix for runtime typing errors

nlaurie:
As Indicated a few days ago ,  our team looked into the runtime typing issues we were getting with 4.0 and believe we came up with a good solution for handling situations where the user has a complex class hierarchy and wants to leverage the query nature of Dexie for their classes but also 100% control no assumptions on the serialization Type back to the db.  

As opposed to Dexie typing to figure out what the user wants in these cases we Added the ability to supply a secondary Type TEntity to take precedence for the POJO where needed for updates etc.  This allows the user of the library to directly specify the serialization type and get all the runtime type-safety

All tests are passed and the change is backward compatible.  There however is still one issue still remaining with the IDType coercion.  We understand the idea behind extracting the 'key' as string to infer the IndexableType from the T , however the typing system gets a bit confused and there are some instances where resolutions go recursive.  It's definitely a complex process switching types around.  Maybe there is a cleaner approach by leaving it more 3.2 ish and give the user a helper function to access the feature if desired, so that keeps the assumptions to a minimum (and easier to maintain)


nlaurie:
Closing Pull Request to Add Fixes for dexie-cloud-addon

PR #1632, opened 2022-10-21
User serialization

nlaurie:
As Indicated a few days ago , our team looked into the runtime typing issues we were getting with 4.0 and believe we came up with a good solution for handling situations where the user has a complex class hierarchy and wants to leverage the query nature of Dexie for their classes but also 100% control no assumptions on the serialization Type back to the db.

As opposed to Dexie typing to figure out what the user wants in these cases we Added the ability to supply a secondary Type TEntity to take precedence for the POJO where needed for updates etc. This allows the user of the library to directly specify the serialization type and get all the runtime type-safety

All tests are passed and the change is backward compatible. There however is still one issue still remaining with the IDType coercion. We understand the idea behind extracting the 'key' as string to infer the IndexableType from the T , however the typing system gets a bit confused and there are some instances where resolutions go recursive. It's definitely a complex process switching types around. Maybe there is a cleaner approach by leaving it more 3.2 ish and give the user a helper function to access the feature if desired, so that keeps the assumptions to a minimum (and easier to maintain)

Also added changes to dexie-cloud-addon


src/public/types/collection.d.ts
```diff
@@ -26,20 +26,20 @@ export interface Collection<T=any, TKey=IndexableType> {
...
   toArray(): PromiseExtended<Array<T>>;
   toArray<R>(thenShortcut: ThenShortcut<T[], R>) : PromiseExtended<R>;
   uniqueKeys(): PromiseExtended<IndexableTypeArray>;
   uniqueKeys<R>(thenShortcut: ThenShortcut<IndexableTypeArray, R>): PromiseExtended<R>;
-  until(filter: (value: T) => boolean, includeStopEntry?: boolean): Collection<T, TKey>;
+  until(filter: (value: T) => boolean, includeStopEntry?: boolean): Collection<T, TKey, TEntity>;
   // Mutating methods
   delete(): PromiseExtended<number>;
   modify(changeCallback: (obj: T, ctx:{value: T}) => void | boolean): PromiseExtended<number>;
-  modify(changes: UpdateSpec<T>): PromiseExtended<number>;
```
dfahlander: Could we keep UpdateSpec<T> here to get better typings for dotted keypaths? I saw it's kept in table.d.ts so I suppose it could be kept here as well?

nlaurie:
Good Catch,  I missed putting that back.  Would you like me to correct it?

Nick

dfahlander:
While going through the changes in the PR, I'm trying to get a grip on what issue you had, in order to allow a solution that would work in your case. Could the issue have been the complex `InsertType` generic failed with your types?

My intention with dexie 4 was to keep it backward compatible (`Table<T, TKey>` should still work as before, but if TKey was an exact string, it would point out the primary key property). I agree it was too complex but I'm looking at a solution where the second parameter could be wither the type (TKey) or a union of optional insert properties (such as `'id' | 'realmId' | 'owner'`). I wonder why this approach wouldn't work and wonder if you have an example where this failed in dexie@4.0.0-alpha.4.

nlaurie:
Yea it was an obscure runtime typing error.  I don't have an isolated
project that showcases it , but if you would like to see what happens in
our project I'm happy to show you.
I have a lot of things going on in my project with types, mixins, Vue
compilation etc.  I spent a good day and a half trying to get it to work
and had to step away until I could understand it better.

I narrowed it down to two general areas

1) (Fixed)
InsertType
When applying the Insert Type to Highly Structured Clases ,  there is
typically more than just the methods that need to be omitted.  Usually
there are support properties and object references to other classes etc.
Cleaner to just specify the serialization contract Hence TEntity (Gives the
consumer full control no assumptions).  I believe the way I did it was
backwards compatible with your idea of the simple entity class since the
TEntity will default to the first generic T if not specified. And
InsertType is still being used as designed, I just omitted the Optional
generic. (assumed you would remove from the InsertType if you liked the
approach)  Runtime Typescript Errors went away when passing a simpler type
to InsertType.  I didn't want to touch InsertType cause if you squint the
wrong way with those complex type structures the logic gets messed up LOL.

2)
export type IDType<T, TKey> = TKey extends keyof T ? T[TKey] : TKey
and
keypaths.d.ts KeyPaths<T>

This one was obscure

In isolation and simplified the IDType performed as expected, but when the
IDType was applied to our Class Types the typescript compiler went haywire
and indicated circular references and type incompatibilities.
It got me scratching my head , theoretically  passing a type into Dexie's
generics should only have resulted in new types and not generated runtime
compilation type errors in our project.  My best guess
is something caused the type model to get polluted or corrupt.  I was going
to revisit it, but not before discussing it with you.

Nick

nlaurie:
Check out PullRequest #1635

dfahlander:
Thanks!

I made a new PR #1636 where I revert to `Table<T, TKey>` but with the 3rd optional property `TInsertType=T`

```ts
class Table<T = any, TKey = IndexableType, TInsertType=T> {
  ...
}
```

It seems hard to cover possible problems caused by a too complex typing system.

Instead offer an alternate generic

```ts
export type EntityTable<T, TKeyPropName extends keyof T, TInsertType = InsertType<T, TKeyPropName>> =
  Table<T, IDType<T, TKeyPropName>, TInsertType>;
```

dfahlander:
PR #1636 is based on master to not leave unintended changes by accident, but with inspiration and selected parts from your PRs. Please let me know if I missed something that is important to make it work in your setup.

PR #1685, opened 2023-03-10
Fix transaction typings

lvivski:
Fixes: https://github.com/dexie/Dexie.js/pull/1624#issuecomment-1464486384 while still preserving correct type for the hook


dfahlander:
Thanks!

PR #1694, opened 2023-03-30
Lambdatest README changes

dfahlander:
This PR is a proposal for README changes when moving over from travis/browserstack to Github Actions and Lambdatest.


dfahlander:
Awaiting to get lambdatests working before we can merge this change.

PR #1717, opened 2023-05-11
Dexie cloud fixes

dfahlander:


b-straub:
I think wherever the underlying observable is an ObservableWithCurrentValue created by createSharedValueObservable (Permissions, Invites, Roles) such an extended Observable should be the type of the relevant property in https://github.com/dexie/Dexie.js/blob/1200ed46899dd6f158280c770abaca1fb3863411/addons/dexie-cloud/src/DexieCloudAPI.ts
this would make the creation of hooks e.g. one similar to UsePermissions more type safe.

dfahlander:
> I think wherever the underlying observable is an ObservableWithCurrentValue created by createSharedValueObservable (Permissions, Invites, Roles) such an extended Observable should be the type of the relevant property in 'DexieCloudAPI.ts'. Since this pull request changes Observable to Rx.Observable it should use the relevant type extension instead, this would make the creation of hooks e.g. one similar to UsePermissions more type safe.

Thanks for the feedback. The previous version did also return Rx observables from dexie cloud API (but liveQuery() still return a custom observable that is compatible with Rx). So we can't return Rx types from dexie or dexie-react-hooks because I wouldn't want dexie or dexie-react-hooks to be dependent of Rx. useObservable() expects a plain observable that may optionally have a getValue() method - as is the case with BehaviorSubject. I don't know exactly how to improve the typings in dexie-react-hooks without introducting a dependency though.

PR #1743, opened 2023-06-14
Undo PR #1459 as it seem to complicate things.

dfahlander:
Aborting transactions that aren't listened to can generate unhandled rejections and make applications less stable. It may also trigger browser bugs and complicate troubleshooting.

This commit undos the changes made in PR #1459 and allows transactions of unsubscribed liveQuery to continue their job.

We can then also skip the special treatment of cached promises that needed to generate new transaction to work around this when a new liveQuery tries to reuse the result of an aborted one.


frankleng:
@dfahlander was wondering why we were getting a ton of aborted transaction errors. thx for fixing

Issue #1734, opened 2023-06-03
TypeError: Cannot convert undefined or null to object in hooksMiddleware


ethan-bernstein:
After deploying a feature using Dexie to tens of thousands of our users, we are seeing devices hit an error in the hooks middleware under `bulkPut`. The devices which hit the error seem to get "stuck" hitting the same error each time our business logic runs.

We haven't been able to find a repro and haven't seen it on device from the team where we could potentially debug further.

From the error and callstack, existingValue is null at the call to `getObjectDiff ` in `hooksMiddleware` in the `.then` of the `getExistingValues` call. And that leads to 'TypeError: Cannot convert undefined or null to object' in the `Object.keys` call at the start of `getObjectDiff`.

Does the check for create vs. update in `hooksMiddleware` need to also check `existingValue` for null where it currently checks for undefined?

ethan-bernstein:
We believe we are hitting this open issue which results in a null object: https://github.com/dexie/Dexie.js/issues/541.

But then doing a subsequent table.put to try to update that object throws 'Cannot convert undefined or null to object.'

Issue #541, opened 2017-05-24
Replacing via Table.put() can result in null value


gregor:
I have no exact steps to reproduce this but again and again we see this issue with different tables in our app:

In a few tables we rely on replacing full items instead of granularly updating them for convenience. According to the docs this is a supported procedure as an exisiting item in the table with the same primary will be replaced by this. It seems like this is not error prone though. Under some circumstances the item is not replaced by the new one. Instead you end up with a table entry for the primary key with a null value. This obviously leads to a number of data inconsistencies.

dfahlander:
What method is being used? Collection.modify() or Table.put() ?
What OS/Browser?
If you use Collection.modify(callbackFn), and set this.value = somethingNew, make sure that this "somethingNew" var is not changed over time. For example, if you set it to a closure var from a for loop, the var could have changed when your callback is called. So make sure to set this.value = somethingThatCannotChangeOverTime.

gregor:
We use Table.put(). One code path, that to my understanding should never result in a null within the database but where it shows:

https://github.com/wireapp/wire-webapp/blob/dev/app/script/event/NotificationService.js#L101-L109
````
 /**
   * Save last notifications id from storage.
   * @param {string} notification_id - Notification ID to be stored
   * @returns {Promise} Resolves with the stored record
   */
  save_last_notification_id_to_db(notification_id) {
    return this.storage_service.save(z.storage.StorageService.OBJECT_STORE.AMPLIFY, NotificationService.CONFIG.PRIMARY_KEY_LAST_NOTIFICATION, {value: notification_id});
  }
};
````
We always forward an object of `{value: NEW_VALUE}`. This can never be null. The *worst* that can happen is `{value: undefined}` meaning `{}` be passed on to:

https://github.com/wireapp/wire-webapp/blob/dev/app/script/storage/StorageService.js#L393-L407

````
  /**
   * Saves objects in the local database.
   *
   * @param {string} store_name - Name of object store where to save the object
   * @param {string} primary_key - Primary key which should be used to store the object
   * @param {Object} entity - Data to store in object store
   * @returns {Promise} Resolves with the primary key of the persisted object
   */
  save(store_name, primary_key, entity) {
    return this.db[store_name].put(entity, primary_key)
      .catch((error) => {
        this.logger.error(`Failed to put '${primary_key}' into store '${store_name}'`, error);
        throw error;
      });
  }
````

dfahlander:
[...]

PR #1770, opened 2023-07-28
Fix null object recovery

ethan-bernstein:
Fix issue #1734 

If a table has a null object, perhaps because of open issue #541, then table.put on that object hits and error 'Cannot convert undefined or null to object' in hooksMiddleware, preventing the client easily recovering from the null object.

The fix is to treat updating a null object as object creation in hooksMiddleware.

### DIFF
--- a/src/hooks/hooks-middleware.ts
+++ b/src/hooks/hooks-middleware.ts
@@ -71,7 +71,7 @@ export const hooksMiddleware: Middleware<DBCore>  = {
                 if (req.type === 'delete') {
                   // delete operation
                   deleting.fire.call(ctx, key, existingValue, dxTrans);
-                } else if (req.type === 'add' || existingValue === undefined) {
+                } else if (req.type === 'add' || existingValue == undefined) {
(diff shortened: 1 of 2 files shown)


dfahlander:
Thanks for this PR! But shouldn't the test expect an update second time? Null is actually added to the key/value table and second put is actually an update from null to `{}`? Maybe the bug should rather be fixed in the else clause where it shouldn't do an object diff if existing value is null?

ethan-bernstein:
@dfahlander I'm not sure. I think having an update without an object diff would also be unexpected if the null was due to a bug (like #541) rather than intentional. In our case, we would want to detect "update without a diff" and treat it like a create. I suspect that might be common, though it's of course hard to anticipate what logic clients have in their hooks.

In practice, it would cause errors in our current updating hook - so returning an update here without an object diff would in some sense be a breaking change.

Let me know if you still prefer to have update without a diff here and I can work up that fix instead.

dfahlander:
Closing. Feel free to reopen.

PR #1795, opened 2023-08-28
Bugfix: Support single-valued compound keys

dfahlander:
Javascript is a bit unsuprisingly forgiving when passing array of string instead of a string:
```js
const obj = {id: 1};
obj.hasOwnProperty("id"); // returns true as expected.
obj.hasOwnProperty(["id"]); // Also returns true (!)

obj["id"] // returns 1 as expected.
obj[["id"]] // also returns 1 (!)

```
This commit makes sure to check typeof keyPath before passing it to getOwn() (alias for `(x, a) => Object.prototype.getOwnProperty.call(x, a))``


dfahlander:
This PR is for dexie master-3 but can be cherry-picked for master as well.

Issue #1651, opened 2022-12-27
Table 'add' method for an auto-incrementing primary key in TypeScript seems incorrect


JJosephttg:
When adding a new item to the table, it seems like if you have an autoincrementing primary key, you shouldn't have to specify that key for the type to add. Currently I am forced to either make the primary key field optional for the type but this is incorrect since when I get the item, it is guaranteed to have a primary key. The add method should take a slightly different type that allows for the primary key not to be specified under the condition that it is autoincrementing.

I get a task like this is probably not easy when it comes to typing, however, a possible solution is to include a 3rd type which either takes an autoincrementing property name, or a type that matches what would actually be added to the database before prepopulating the primary key.

In my example, I am using an id field that is forced to be optional because of the fact that I need to access the ID after its been incremented. However, during adding of the item, there is no need to specify the ID.

In the meantime, is there an alternative solution to this? Currently I just made the id field optional, but I feel like there has to be a better way, since after I get the object back, it isn't optional anymore. I am also hoping not to force type cast it, but if that is the only way I guess that is what I will have to do. And if I specify a generic id when I go to add the item, wouldn't that actually try to add it with the id and fail if there is a dupe?

tomasr8:
I ran into the same issue and just wanted to give an example using the [official docs](https://dexie.org/docs/Typescript):

```ts
class MyAppDatabase extends Dexie {
    contacts!: Dexie.Table<IContact, number>; // number = type of the primkey

    constructor () {
        super("MyAppDatabase");
        this.version(1).stores({
            contacts: '++id, first, last'
        });
    }
}

interface IContact {
    id?: number,
    first: string,
    last: string
}
```
`Icontact.id` is optional here to make e.g. `db.contacts.add({first, last})`  work. The problem appears when using get():

```ts
const contact = db.contacts.get(1)
contact.id // type is number | undefined
```
I'm using non-null assertions to get around this (i.e. `contact.id!`) but it would be great if there was another way..

dfahlander:
The reason is that id is declared as an optional property. contact.id can be undefined. I agree it would be nicer if it would be possible to declare IContact like this:

```ts
interface IContact {
    id: number,
    first: string,
    last: string
}
```

The only reason for not doing so is that id is still optional when calling `db.contacts.add({first: "foo", last: "bar"})`. That line would complain about property 'id' is missing.

## Dexie 4's TInsertType

Therefore, there's a new declaration style in dexie@4 that can optionally be used:

```ts
[...]

Issue #1358, opened 2021-07-08
Why nullable ids in typescript?


Ryunaq:
Is there any reason why the `ids` of the interface of a model should be nullable? I was reading this example https://github.com/dfahlander/dexie-todo-list/blob/master/models/TodoItem.ts and saw you made them nullable.
However from my understanding by defining `'++id'`  on the `.stores` function `id` should always exist. Am I missing something?

Ryunaq:
Nevermind its so it actually auto increments when adding an element, but then that forces us to check for the availability of the `id` constantly despite being assured it actually exists.

PR #1810, opened 2023-10-05
Expose TInsertType of the Table interface

tomasr8:
Related issues: #1651 #1358 

With this change we can specify separate types for the get()/add() methods.
This avoids having to make the primary key optional which requires non-null checks/assertions
anywhere the result of get() is used.

```ts
class MyAppDatabase extends Dexie {
    contacts!: Dexie.Table<IContact, number, Omit<IContact, 'id'>>; // <-- specify Insert type

    constructor () {
        super("MyAppDatabase");
        this.version(1).stores({
            contacts: '++id, first, last'
        });
    }
}

interface IContact {
    id: number, // <-- not nullable anymore
    first: string,
    last: string
}

db.contacts.add({first, last}) // same as before
const contact = await db.contacts.get(1)
contact.id // <-- type is now number rather than number | undefined
```

### DIFF
--- a/src/public/index.d.ts
+++ b/src/public/index.d.ts
@@ -32,7 +32,7 @@ export { UpdateSpec } from './types/update-spec';
 export * from './types/insert-type';
 
 // Alias of Table and Collection in order to be able to refer them from module below...
-interface _Table<T, TKey> extends Table<T, TKey> {}
+interface _Table<T, TKey, TInsertType> extends Table<T, TKey, TInsertType> {}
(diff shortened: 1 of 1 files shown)


dfahlander:
I see, this seems to be related to using the `Dexie.Table<>` type. I see it's referenced in the docs (for historical reasons). I believe the docs should be updated to `import { Table } from 'dexie'` instead, as that works in both dexie@3 and dexie@4. Thanks for the PR also. I believe it wouldn't hurt to merge it as well. Need just to verify a case for it in test-typings.ts also.

PR #1814, opened 2023-10-17
#REFACTOR: remove useless import

wsafight:
remove useless import


dfahlander:
Thanks for the hint. Rollup's tree-shaking is taking care of this but it's nice to get rid of them anyway.
However, I prefer to view this PR as an issue and not merging it directly as it is safer to redo refactoring things in isolation. Keeping the PR open as a reminder.

Issue #1842, opened 2023-12-10
Unnecessary properties in primKey and indexes in schema


Stretsh:
TLDR
primKey is always considered unique and cannot be multi-value. Indexes cannot be marked as auto-increment. Why leave those properties in there?

Background
I'm trying to dynamically create objectStores/tables (Storage app, each warehouse needs their own table, added through the UI). I've been playing around with the sample code in #988. I added a `console.log(db.tables)` at the end of the `playAround()` function.

When I add a new entry in that function when calling the `changeSchema` function, using `&id`, the newly created table shows primKey with a `unique: true` property. Expected behavior, the Dexie newbie in me thought.

Adding a new entry and running it again, now shows both tables. However, `unique` for the primKey of the first table is now set to `false`, while for the new table is `true`. Rinse and repeat, and all older tables have it set to false, while only the newly created table contains the property set true.

This kinda confused and worried me, so I spent all morning trying to understand what is happening here.

Turns out that it's explicitly set to false in line 315 here https://github.com/dexie/Dexie.js/blob/f088a7df990b9eb697f73e98603a01e4fe95dd65/src/classes/version/schema-helpers.ts#L312-L320 when `db.open()` is called (via `readGlobalSchema`).

In contrast, when creating the store using `db.version().stores()`, it just takes the first of the indexes and converts that to the primKey, only checking if `primKey.multi` is true. See https://github.com/dexie/Dexie.js/blob/f088a7df990b9eb697f73e98603a01e4fe95dd65/src/classes/version/version.ts#L25-L38
I believe the behavior should be consistent: either show unique as false everywhere, or show it as true. Or... remove it completely for the primKey, since its uniqueness is implied.

And while we're at it: since only the primKey is allowed to have auto set, why not remove it from the indexes? I'm thinking the same for primKey and multi-value.

IMHO, as it is now, it can cause unnecessary confusion for (especially new) developers looking into Dexie.

dfahlander:
Primary key is also an index and there are reasons to reuse much of the code between the two. When dexie reverse-engineers the schema, it might be setting the orimKey's uniqueness flag to false in error. However, this does not affect the behavior since the primary key is always unique.

So basically this is a bug but IMHO not of certian high prio. A PR is welcome to change line 315 in schema-helpers and make other adjustments needed for unit tests to pass (if nescessary).

Stretsh:
My PR-fu is very weak, virtually non-existent, but I might give it a shot.

PR #1852, opened 2023-12-15
Sets the unique flag for primKey to true

Stretsh:
Although uniqueness is implied for the primary key, having this flag set to false can cause confusion. As discussed in #1842.


dfahlander:
We just need a unit test for it before merging it - somewhere in tests-open.js. We should verify as discussed in #1842 that if we create a database (providing a schema) and then close it and reopen it dynamically, the primary key is marked as unique.

Stretsh:
Good that I needed to add the tests, cause I discovered somewhere else that needed to be changed. If one didn't include an `&` when defining the schema, it would still set the `unique` flag to false when creating the table.

Hope it's good now. I know it's simple stuff for you fellas with strong PR-fu, but I'm kinda excited about my first PR in a very long time!

dfahlander:
Very nice test! This PR was really well done.

PR #1896, opened 2024-02-05
Added SkipTables Option

nlaurie:
In my effort to properly convert dexie-encrypted to work with more complex scenarios,  I found the need to completely exclude tables from the import export process.  Mainly the encryption metadata.  

So I added the options into the project both import and export,  to simply skip the requested list of tables in all operations.  

I feel this is a good addition to round out the options for controlling the backups. 

It's backwards compatible and does nothing if not specified. 

Regards
Nick

### DIFF
--- a/addons/dexie-export-import/src/export.ts
+++ b/addons/dexie-export-import/src/export.ts
@@ -5,6 +5,7 @@ import { DexieExportedTable, DexieExportJsonStructure } from './json-structure';
 import { TSON } from './tson';
 
 export interface ExportOptions {
+  skipTables?: string[],
   noTransaction?: boolean;
   numRowsPerChunk?: number;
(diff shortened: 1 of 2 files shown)


dfahlander:
Would the filter option not be enough to achieve the same as the proposed `skipTables` option? Or would it still fail with dexie-encrypted if so?

nlaurie:
I started with the filter ,  it mostly worked but ended up colliding with the clearTablesBeforeImport. 

When importing into a database set up with encryption. you don't want to loose you metadata nor do you want to import/export it.  so I set skipTables: ['_encryptionSettings'].  This way the backup is encryption Free and can be easily imported into a new DB possibly with different encryption params and it will reimport all the data perfectly.

PR #1909, opened 2024-02-28
Allow setting chunk size per table

lvivski:
Single Database can have object stores with varying complexity and amount of data in them. Having one `modifyChunkSize` value may not be sufficient for all scenarios. 

This PR introduces an option to set `modifyChunkSize` per table.


dfahlander:
Sorry for taking time. Yes I merged it.

lvivski:
Thanks!

PR #1973, opened 2024-04-29
Allow the user to import into a different database than the one specified in the file

tobiasBora:
Fix https://github.com/dexie/Dexie.js/issues/1972 The callback allows a quite flexible solution as I can ask to the user a different name if the database already exists etc…


dfahlander:
Could we use a simpler approach and just allow an optional `name` property? I suppose letting the name depend on the other info from the file is less common use case but when that is needed, one could use [peakImportFile()](https://dexie.org/docs/ExportImport/dexie-export-import#usage) to get that information beforehand.

tobiasBora:
Sorry, I missed your message. I wasn't aware of `peakImportFile`, it seems indeed to be good enough for my need if an optional `name` is provided. I'll try to see if it works.

tobiasBora:
Ok, this one is tested (and fixed) now that I fixed my issue with https://github.com/dexie/Dexie.js/issues/1995#issuecomment-2166951977

tobiasBora:
Any update on this?

Issue #1995, opened 2024-05-15
"default" is not exported by "../../../Dexie.js/dist/dexie.min.js", imported by "../../../Dexie.js/import-wrapper-prod.mjs".


nlaurie:
When building via Vite.  Getting 

`
x Build failed in 8.15s
error during build:
RollupError: [vite-plugin-pwa:build] ../../../Dexie.js/import-wrapper-prod.mjs (4:7): "default" is not exported by "../../../Dexie.js/dist/dexie.min.js", imported by "../../../Dexie.js/import-wrapper-prod.mjs".
file: /Users/nlaurie/Development/metafig/Dexie.js/import-wrapper-prod.mjs:4:7
`

Everything go back to building fine when I revert to a previous package.json that looks like this

`
      "production": {
        "module": "./import-wrapper-prod.mjs",
        "import": "./import-wrapper-prod.mjs",
        "require": "./dist/dexie.min.js",
        "default": "./dist/dexie.min.js"
      },
`

dfahlander:
How does the non-working package.json look? When I look at current one, it has

```
      "production": {
        "module": "./import-wrapper-prod.mjs",
        "import": "./import-wrapper-prod.mjs",
        "require": "./dist/dexie.min.js",
        "default": "./dist/dexie.min.js"
      },
```

Seems identical?

nlaurie:
Left Non Working,  Right Working

![image](https://github.com/dexie/Dexie.js/assets/10317679/fd2dbf1c-8cc0-47c1-9178-d5b61cae1c25)

nlaurie:
It was a couple months back when the issue cropped in when the wrapper was introduced, I figured it would come up and get fixed with the release of 4.  I went through our code and made sure all the imports were consistent.  since I've seen this sort of error with libraries that dont have default exports and not using the * notation for the import.  import * as xyx from 'lib',  but Dexie has default exports.   so were using      import Dexie from 'dexie'

dfahlander:
Tried to reproduce just now: https://github.com/dfahlander/dexie-vite-starter
I get it working with a clean vite project with typescript / react and adding dexie and using dexie-react-hooks,

But I saw there was an issue in vite https://github.com/vitejs/vite/issues/2679

Which version of vite are you using?

nlaurie:
Vite 5.2.11

I switched the wrappers to , and its building with the configuration.  Shouldn't the import and browser be using the modern ESM build?
[...]

PR #1997, opened 2024-05-20
export ./dist/dexie.mjs

KittenApps:
Makes importing `dexie/dist/dexie.mjs` not restricted by package.json `exports` field.

Refs #1995


dfahlander:
Thanks!

Issue #1976, opened 2024-05-03
Transaction.table() is missing generic overload with insert type


SuNNjek:
With Dexie 4 there's now an option to specify an insert type for tables for optional auto-increment properties.

The `Transaction.table()` function is currently still missing an overload where you can specify an insert type though:
https://github.com/dexie/Dexie.js/blob/b2fd7c0df1da4145f39493419557c5fcc244447b/src/public/types/transaction.d.ts#L15-L17

It would be great if this could be added.

PR #1998, opened 2024-05-24
TInsertType in table() methods on Dexie and Transaction

dfahlander:


dfahlander:
This PR is to resolve #1976

Issue #1926, opened 2024-03-20
Rich-text realtime collaboration


dfahlander:
Add support for realtime communication of text editors.

See what we can use from tiptap and hocuspocus server to integrate it seamlessly in dexie cloud.

dusty-phillips:
That would be such a killer feature.

dfahlander:
I'm looking into integrating Y.js into Dexie.js, not only Dexie Cloud. So this issue will be twofold:

## Dexie.js ❤️ Y.js

The plan is that Dexie.js will support Y.Docs as properties on objects and thus be a storage provider of Y.js in itself even without any addon required. If it turns out to be too much of added code (which I do not think it will be), we might lift this out to an addon instead, but since Y.js ecosystem is so promising and leading, it feels more right to support it natively.

Dexie's role in the Y.js ecosystem will be similary to y-indexeddb but allowing docs to be organised into props of objects and queried just like other props of Dexie tables. Allow tagging docs and index them and reference them from other tables. Also the lifetime of the Y.Doc instances will be cached and handled, taking away the burden for developers to maintain and organize Y.Doc instances.

## Dexie Cloud ❤️ Y.js

* Dexie Cloud will handle Y.Docs seamlessly and sync, support realtime editing of text using its existing connection and access control features.
* Dexie Cloud will also support Y.js awareness protocol for seeing other user's cursors etc in text documents.

dfahlander:
Update

Dexie + Y.js support is almost done. Kevin (Y.js author) has been very helpful with actively supporting me in API questions. The dexie-cloud support is still to be implemented but the design for it is done.

scottrblock:
This is excellent @dfahlander! Will it eventually be possible to use Dexie Cloud REST API to update a Y.js doc? My use case would be from a node.js or similar server/endpoint.

Tangential, but I am assuming REST API is best way to access/update records in Dexie Cloud from a backend because the npm package would rely on indexdb being available, but not sure.

dfahlander:
I think we'd need to collect the requirements on REST API for sync use cases together to find a solution. The /sync endpoint is possible to call from a server in theory but there are lots of undocumented data formats etc that would need to be documented. Supposely we could add a simplified sync api that is targeted for node clients rather than dexie-cloud-addon on web.

Adding a REST endpoint for Y-updates would not be a problem though, just want to think it through to get it right.

scflode:
[...]

PR #2016, opened 2024-06-19
Support Y.js

dfahlander:
See #1926


dfahlander:
Replaced by #2045 that contains both the dexie and dexie-cloud-addon changes

PR #2049, opened 2024-08-02
Update update-spec.d.ts

janmarques:
Hi,

The issue https://github.com/dexie/Dexie.js/issues/2001 can be resolved with this change.

I have only tested this by changing it by messing with the `node_modules\.pnpm\dexie@4.0.8\node_modules\dexie\dist\dexie.d.ts` file in my node_modules, resolving the typescript error.

Using typescript 5.5.4, and dexie 4.0.8.

### DIFF
--- a/src/public/types/update-spec.d.ts
+++ b/src/public/types/update-spec.d.ts
@@ -1,4 +1,4 @@
 import { KeyPaths, KeyPathValue } from "./keypaths";
 import { PropModification } from "./prop-modification";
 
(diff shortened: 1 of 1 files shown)


janmarques:
@dfahlander wanted to make sure you saw this. Take your time of course.

dfahlander:
Seems to be a typo in the spec. Remove equals sign between interface and leading bracket?

image

janmarques:
Yes you are right. 
I digged it up again and it is without the `=`, and my client code runs without warning.

PR #2074, opened 2024-09-21
Uses `cmp` for sortBy

### LINKED ISSUE #2073: `Collection.sortBy` doesn't use `IndexedDB.cmp` for Uint8Array values
`Collection.sortBy` function uses `<` or `>` for comparing values for the sort.
https://github.com/dexie/Dexie.js/blob/e41f017b4fb0ddf23d037c44090acc5184b41135/src/classes/collection/collection.ts#L153

It causes unexpected behaviour for Typed arrays such as Uint8Array.
For example, it sorts the values in lexicographical order like this.
```js
new Uint8Array([1,46]) > new Uint8Array([1,146])
> true
```
This is inconsistent to the treatment of the Typed arrays in the IndexedDB.
We have to use the `IndexedDB.cmp` for comparing function.

genki:
Fixes #2073

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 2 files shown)


dfahlander:
Thanks! Good you saw this and fixed. Will however change to use the custom cmp from cmp.ts (because it is faster and doesn't throw when data contains null keys)

genki:
@dfahlander Okay, I did it :)

dfahlander:
Thanks for the update. Could you revert the updated lockfile though? It will generate conflicts with another PR I have that also updates it. It might also be a reason for the tests failing.

genki:
@dfahlander I see. Just reverted.

Issue #2064, opened 2024-09-02
Error in Dexie 4: Failed to execute 'cmp' on 'IDBFactory': The parameter is not a valid key


probert94:
I recently updated Dexie from 3.2.6 to 4.0.8. Since then one of my queries throws the error `Failed to execute 'cmp' on 'IDBFactory': The parameter is not a valid key` in certain cases.
The query looks like this:
```
await this.table("table")
    .where({foreignKey1: "1", foreignKey2: "1"})
    .toArray()
```
After some investigation I found, that the error is thrown, if there are entries for "foreignKey1" where "foreignKey2" is "null".
In my case the foreignKey1 is never "null" so switching order of the fields in "where" solves the error and this code works:
```
await this.table("table")
    .where({foreignKey2: "1", foreignKey1: "1"})
    .toArray()
```
As I did not define a compound index for the two fields it shows a warning but returns the correct results.

dfahlander:
Thanks. The behaviour should be fixed by switching to using [cmp.ts](https://github.com/dexie/Dexie.js/blob/master/src/functions/cmp.ts) instead of idb.cmp() at [this location](https://github.com/dexie/Dexie.js/blob/master/src/classes/table/table.ts#L151). By using cmp.ts (which is also much faster that indexedDB.cmp()), a null comparision returns NaN instead of throwing, which is what would be correct here.

dfahlander:
Resolved in #2076

Nedi11:
When will the package be updated with this fix

dfahlander:
A new release of dexie will probably be out tomorrow with the latest from master.

PR #2076, opened 2024-09-26
Cmp instead of idb

ankitjha420:
This PR aims to fix issue #2064 
What I've changed is basically the comparison function, from idb.cmp() to cmp() defined at src/functions/cmp.ts

### DIFF
--- a/src/classes/table/table.ts
+++ b/src/classes/table/table.ts
@@ -148,7 +148,7 @@ export class Table implements ITable<any, IndexableType> {
     const idb = this.db._deps.indexedDB;
 
     function equals (a, b) {
-      return idb.cmp(a,b) === 0; // Works with all indexable types including binary keys.
+      return cmp(a, b); // Works with all indexable types including binary keys.
     }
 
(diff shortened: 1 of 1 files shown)


src/classes/table/table.ts
```diff
@@ -146,9 +146,8 @@ export class Table implements ITable<any, IndexableType> {
     // and filter the rest.
     const { idxByName } = this.schema;
     const idb = this.db._deps.indexedDB;
-
     function equals (a, b) {
-      return idb.cmp(a,b) === 0; // Works with all indexable types including binary keys.
+      return cmp(a, b); // Works with all indexable types including binary keys.
```
dfahlander: You mean `cmp(a, b) === 0;` ?

ankitjha420: dumb mistake on my part, fixed it. sorry about that

dfahlander:
Thanks for the fix. I think this was the last place where we are using idb.cmp. If you have time, could you check if we could also get rid of the line `const idb = this.db._deps.indexedDB;`. I believe it's not needed anymore.

ankitjha420:
I've checked it out and it's not being used anywhere else in table.ts, so removed it

Issue #1954, opened 2024-04-06
Feature Request: add direction to sortBy() method


tacman:
Instead of
   
     db.friends.orderBy('lastName').reverse().offset(10).limit(5);

I'd like to do

     db.friends.orderBy('lastName', 'desc').offset(10).limit(5);

That is, if I'm calling orderBy I'd like to optionally add the direction without having a separate statement.

dfahlander:
Not sure whether it's the best signature for this as `orderBy('property', 'desc')` could be interpreted as if 'desc' was a property to order on secondly. I think we could wait with this as there's already another possibility to orderBy in descending order using `orderBy('property').desc()` (which is deprecated by the way to the favour of `orderBy('property').reverse()`.

PR #2077, opened 2024-09-26
Feature Added: orderBy() direction parameter

ankitjha420:
This PR is for issue #1954 (don't know if its really important)

I've added an additional parameter 'order' which can either be 'asc' or 'desc' (its 'asc' by default) according to which the result is/is not reversed. Also updated the function signature in table.d.ts to reflect this change. Lastly I've added additional test for this new feature in test/tests-table.js. Hope this helps

PS: Please let me know if there's something wrong I'm new to this, thx

### DIFF
--- a/src/classes/table/table.ts
+++ b/src/classes/table/table.ts
@@ -244,11 +244,12 @@ export class Table implements ITable<any, IndexableType> {
    * https://dexie.org/docs/Table/Table.orderBy()
    * 
    **/
-  orderBy(index: string | string[]) {
-    return new this.db.Collection(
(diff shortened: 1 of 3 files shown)


dfahlander:
Not sure whether it's the best signature for this as `orderBy('property', 'desc')` could be interpreted as if 'desc' was a property to order on secondly. I think we could wait with this as there's already another possibility to orderBy in descending order using `orderBy('property').desc()` (which is deprecated by the way to the favour of `orderBy('property').reverse()`.

I will close this and the original feature request.

PR #2086, opened 2024-10-21
Advanced typings for a strongly typed stores spec

dfahlander:


dfahlander:
This PR can probably be closed but kept as a reference to what could be doable. The problem with this could be that it uses template strings and conditionals too heavily, making it prone to strange typescript issues for the users.

Issue #2075, opened 2024-09-23
Is there a plan to make liveQuery more compatible with Svelte5 runes


newholder:
With the introduction of runes, store is kinda obsolete. Is there any plan to support runes once svelte5 released?

I use it like this in svelte5

```svelte
let _friends = liveQuery(async () => {
      return await db.friends
        .where("age")
        .between(18, 65)
        .toArray();
    });

let friends = $derived($_friends)
```

dfahlander:
Thanks for bringing it up. Also asked on stackoverflow: https://stackoverflow.com/questions/78089371/dexie-livequery-with-svelte-5-runes

Need more knowledge about Svelte 5 and runes before I could do anything about it but a PR would be much welcome. Could we introduce some kind of svelte helper library that delivers runes for live queries? A bit like we have for react?

molarmanful:
The recently-added [fromStore](https://svelte-5-preview.vercel.app/docs/imports#svelte-store-fromstore) alongside the wrapper from #1907 worked for me:

```js
import { liveQuery } from 'dexie'
import { fromStore } from 'svelte/store'

export const liveQ = q => fromStore({
  subscribe: (run, invalidate) =>
    liveQuery(q).subscribe(run, invalidate).unsubscribe,
})

// usage

const friends = liveQ(async () => await db.friends.where('age').between(18, 65).toArray())

$inspect(friends.current)
```

ericdudley:
Hello, I am also trying to use Dexie.js in a Svelte5 project with runes. I started by following the existing tutorial [Get started with Dexie in Svelte](https://dexie.org/docs/Tutorial/Svelte) tutorial. However, I got stuck when trying to make a liveQuery reactive to props/state.

I was able to get it working by combining `$derived.by` with a noop reference to the state I want to react to (in this case `namePrefix`). Is this the correct way of parameterizing a live query? Re-creating the liveQuery whenever the parameter changes?

```
<script lang="ts">
	import { db } from '$lib/store';
	import { liveQuery } from 'dexie';

	let { namePrefix }: { namePrefix: string } = $props();

	let friends = $derived.by(() => {
		// noop just to make it reactive
		namePrefix;

		return liveQuery(() => db.friends
				.where('name')
				.startsWithIgnoreCase(namePrefix ?? '')
				.toArray()
		);
	});
</script>
```

dfahlander:
[...]

Issue #2089, opened 2024-10-30
Async Loading Status


ericdudley:
Hello, I am using Dexie.js for the first time with my Svelte5 app, and I am trying to figure out how to show the query async status lifecycle.
```
	let transactions = $derived.by(() => {
		// noop just to make it reactive
		prefix;
		startDate;
		endDate;

		return liveQuery(() =>
			db.tx
				.where('yyyyMMDd')
				.between(format(startDate, 'yyyy-MM-dd'), format(endDate, 'yyyy-MM-dd'), true, true)
				.and((tx) => !!tx.label?.startsWith(prefix ?? ''))
				.sortBy('yyyyMMDd')
		);
	});
```

For the initial load, I can check for `transactions` or `$transactions` being undefined, but after I change the query inputs I would like to show some "loading" UI while the new query is running. Likewise, if an error occurs, I would like to be able to show that.

What is the best practice for showing these intermediate loading + error states?

dfahlander:
liveQuery() does not emit anything until a query finish. If arguments change and you gain a new store, Svelte will keep showing the value from the previous store until the new store emits something. For most cases this is good because it could otherwise have been causing unnecessary flickering. liveQuery emits errors according to the [TC39 Observable proposal](https://github.com/tc39/proposal-observable?tab=readme-ov-file#observable) but Svelte stores does not consume errors.

The following helper should give what you need (not tested):

```ts
export function svelteLiveQuery(querier) {
  return {
    subscribe: (subscriber) => {
      let current = {
        value: null,
        error: null,
        isLoading: true,
      };
      subscriber(current); // immediately emit initial value before loading
      const subscription = liveQuery(querier).subscribe(
        (value) => {
          current.isLoading = false;
          current.error = null;
          current.value = value;
          subscriber(current);
        },
        (error) => {
          current.isLoading = false;
          current.error = error;
          subscriber(current);
        }
      );
      return () => subscription.unsubscribe();
    },
  };
}
```

There might be some even better helper (such as a liveQuery rune or something) that would feel more native to Svelte. I'm not a Svelte expert but any contribution would be welcome.

dfahlander:
*Note: I just updated the my code snippet with type annotations and renamed `subscriber` to `emit`*

PR #2116, opened 2025-01-09
Create a library for a reactive Dexie.js query in Svelte 5

oliverdowling:
I put the code from issue #2075 into a Svelte library.

I just used [`sv create`](https://svelte.dev/docs/cli/sv-create) and stripped it down a bit.

### DIFF
--- a/libs/dexie-svelte-query/README.md
+++ b/libs/dexie-svelte-query/README.md
@@ -1,0 +1,30 @@
+# Dexis Svelte Query
+
+Svelte reactive state query for fetching using Dexie.js
+
+## Usage
+
+```svelte
+<script lang="ts">
+	import { stateQuery } from 'dexie-svelte-query';
+	import { db } from './db';
+
+	let minAge = $state(18);
+	let maxAge = $state(65);
+
+	const friendsQuery = stateQuery(
+		() => db.friends.where('age').between(minAge, maxAge).toArray(),
+		() => [minAge, maxAge]
+	);
+	let friends = $derived(friendsQuery.current);
+</script>
+
+<div>
+	<h2>Friends</h2>
+	<ul>
+		{#each friends ?? [] as friend (friend.id)}
+			<li>{friend.name}, {friend.age}</li>
+		{/each}
+		</ul>
+</div>
+```
--- a/libs/dexie-svelte-query/src/lib/index.ts
+++ b/libs/dexie-svelte-query/src/lib/index.ts
@@ -1,0 +1,1 @@
+export * from "./stateQuery.svelte.js";
--- a/libs/dexie-svelte-query/.gitignore
+++ b/libs/dexie-svelte-query/.gitignore
@@ -1,0 +1,4 @@
+# Output
+/.svelte-kit
+/dist
+
--- a/libs/dexie-svelte-query/.npmrc
+++ b/libs/dexie-svelte-query/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
--- a/libs/dexie-svelte-query/package.json
+++ b/libs/dexie-svelte-query/package.json
@@ -1,0 +1,39 @@
+{
+	"name": "dexie-svelte-query",
+	"version": "1.0.0",
+	"description": "Svelte reactive state query for fetching using Dexie.js",
+	"author": "Oliver Dowling (https://github.com/oliverdowling)",
+	"scripts": {
+		"build": "pnpm run package",
+		"package": "svelte-package && publint",
+		"prepublishOnly": "pnpm run package",
+		"check": "svelte-check --tsconfig ./tsconfig.json",
+		"check:watch": "svelte-check --tsconfig ./tsconfig.json --watch"
+	},
+	"files": [
+		"dist",
+		"!dist/**/*.test.*",
+		"!dist/**/*.spec.*"
+	],
+	"svelte": "./dist/index.js",
+	"types": "./dist/index.d.ts",
+	"type": "module",
+	"exports": {
+		".": {
+			"types": "./dist/index.d.ts",
+			"svelte": "./dist/index.js"
+		}
+	},
+	"peerDependencies": {
+		"dexie": "^3.2 || ^4.0.1",
+		"svelte": "^5.0.0"
+	},
+	"devDependencies": {
+		"@sveltejs/package": "^2.3.7",
+		"dexie": "^4.0.10",
+		"publint": "^0.3.0",
+		"svelte": "^5.17.1",
+		"svelte-check": "^4.1.3",
+		"typescript": "^5.7.3"
+	}
+}
--- a/libs/dexie-svelte-query/src/lib/stateQuery.svelte.ts
+++ b/libs/dexie-svelte-query/src/lib/stateQuery.svelte.ts
@@ -1,0 +1,17 @@
+import { liveQuery } from "dexie";
+
+export function stateQuery<T>(
+	querier: () => T | Promise<T>,
+	dependencies?: () => unknown[]
+): { current?: T } {
+	const query = $state<{ current?: T }>({ current: undefined });
+	$effect(() => {
+		dependencies?.();
+		return liveQuery(querier).subscribe((result) => {
+			if (result !== undefined) {
+				query.current = result;
+			}
+		}).unsubscribe;
+	});
+	return query;
+}
--- a/libs/dexie-svelte-query/tsconfig.json
+++ b/libs/dexie-svelte-query/tsconfig.json
@@ -1,0 +1,14 @@
+{
+	"compilerOptions": {
+		"allowJs": true,
+		"checkJs": true,
+		"esModuleInterop": true,
+		"forceConsistentCasingInFileNames": true,
+		"resolveJsonModule": true,
+		"skipLibCheck": true,
+		"sourceMap": true,
+		"strict": true,
+		"module": "NodeNext",
+		"moduleResolution": "NodeNext"
+	}
+}


dfahlander:
Thanks! The reason it fails to build seem to be outside of this PR - something has become broken elsewhere.

DanConwayDev:
what about this little supplementary helper function?
```typescript

export function liveQueryState<T, I>(
	querier: () => T | Promise<T | undefined>,
	dependencies: () => unknown[] = () => [],
	default_state: I
): T | I {
	return $derived(stateQuery(querier, dependencies).current || default_state);
}
```
with usage:
```typescript
let friendsQuery = liveQueryState(
  () => db.friends.where('age').between(minAge, maxAge).toArray(),
  () => [minAge, maxAge],
  [{name: 'Jesus' age: -Infinity}]
);
```

It returns what svelte users would expect from a svelte version of liveQuery. 

I added the `default_state` feature as its easier to optionally replace undefined here than in each svelte component.

You could add a `default_state` to `stateQuery` but I don't think its needed and I couldn't work out how to make the typings work when with a default value of `undefined`.

```typescript
export function stateQuery<T, I>(
	querier: () => T | Promise<T | undefined>,
	default_state: I,
	dependencies: () => unknown[] = () => []
): { current: T | I } {
	const query = $state<{ current: T | I }>({ current: default_state });
	$effect(() => {
		dependencies?.();
		return liveQuery(querier).subscribe((result: T | undefined) => {
			if (result === undefined) {
				query.current = default_state;
			} else {
				query.current = result as T;
			}
		}).unsubscribe;
	});
	return query;
}
```

oliverdowling:
Part of the issue was that rendered components would flicker very briefly when the dependencies changed. Your example would have the same issue of flickering, but to a default value instead.

I think that a better approach for your use-case would be to not have the check for `undefined` at all and then use it like this:

```typescript
const friendsQuery = liveQueryState(
  () => db.friends.where('age').between(minAge, maxAge).toArray(),
  () => [minAge, maxAge]
);
let friends = $derived(liveQuery.current ?? [{name: 'Jesus', age: -Infinity}]);
```

Keeping the `undefined` check would make it an initial-state rather than a default-state.

oliverdowling:
I was also looking at Issue #2089 and considered something along the lines of:

```typescript
export function stateQuery<T>(querier: () => T | Promise<T>, dependencies?: () => unknown[]) {
	const query = $state<{ result?: T; isLoading: boolean; error?: any }>({
		result: undefined,
		isLoading: true,
		error: undefined,
	});
	$effect(() => {
		dependencies?.();
		return liveQuery(querier).subscribe(
			(result) => {
				query.error = undefined;
				if (result !== undefined) {
					query.result = result;
					query.isLoading = false;
				} else {
					query.isLoading = true;
				}
			},
			(error) => {
				query.error = error;
				query.isLoading = false;
			}
		).unsubscribe;
	});
	return query;
}
```

And then you could achieve a default state with:

```typescript
const friendsQuery = stateQuery(
	() => db.friends.where('age').between(minAge, maxAge).toArray(),
	() => [minAge, maxAge]
);
let friends = $derived(
	friendsQuery.isLoading || friendsQuery.result === undefined
		? [{ name: 'Jesus', age: -Infinity }]
		: friendsQuery.result
);
```

DanConwayDev:
Its annoying that it cant be used with a single line but i guess that's down to the constraints of `$state` and `$derived` svelte.

but as its can't your `isLoading` and `error` properties look useful.

oliverdowling:
Is it just an initial state that you're looking for, or for any time it is loading? How long do your queries take?

Why is your default state a different type to the query result type?

Would this solve your issue:

```typescript
export function stateQuery<T>(querier: () => T | Promise<T>, dependencies?: () => unknown[], initialValue?: T) {
	const query = $state<{ value?: T; isLoading: boolean; error?: any }>({
		value: initialValue,
		isLoading: true,
		error: undefined,
	});
...
const friendsQuery = stateQuery(
	() => db.friends.where('age').between(minAge, maxAge).toArray(),
	() => [minAge, maxAge],
	[{ name: 'Jesus', age: -Infinity }]
);
```

DanConwayDev:
> Is it just an initial state that you're looking for, or for any time it is loading? Svelte's `$state` function just takes an initial value but I was unsure because of re-assigning the value in liveQuery in your example.
>
> Why is your default state a different type to the query result type? If it is intentionally different, then I think it should be separated into a `$derived` rune in the component.

having a different initial and loading type is an edge case and, as you say, it could be implemented  with `$derived` and `isLoading`.

I see you suggested renaming `current` to `result` but I think `current` communicates more meaning:
 `{current?: T, isLoading: boolean, error: unknown}`

I started on this journey because I wanted to upgrade from svelte 4 to svelte 5. I previously was using the pattern:
```typescript
/// $lib/components/things.ts
...
let search_input = writable('');
$: things = query_centre.searchThings($search_input);
...
/// $lib/query_centre.ts
class QueryCentre {
  searchThings(query: string) {
	/// do stuff to find things and update db...
	return liveQuery(async () => {
	  return await db.repos.where('searchWords')
		.startsWithAnyOfIgnoreCase(query)
		.distinct()
		.toArray();
	});
}
const query_centre = new QueryCentre();
export default query_centre;
```
but with svelte 5 I wanted to do this:
```typescript
/// $lib/components/things.ts
...
let search_input = $state('');
let things: T | undefined = $derived(query_centre.searchThings(search_input));
...
/// $lib/ query_centre.ts
class QueryCentre {
  searchThings(query: string) {
	/// do stuff to find things and update db...
	return liveQueryState(
          () => db.repos.where('searchWords').startsWithAnyOfIgnoreCase(query).distinct().toArray(),
          () => [query], // I now realize this is unnecessary 
	});
}
const query_centre = new QueryCentre();
export default query_centre;
```
but I dont think liveQueryState can be implemented to work like that.

I could use your function as is but having to define `things_query` and `things` feels too boilerplatey:
```typescript
/// $lib/components/things.ts
...
let search_input = $state('');
let things_query = $derived(query_centre.searchThings(search_input));
let things = $derived(things_query.current || []);
...
}
/// $lib/ query_centre.ts
class QueryCentre {
  searchThings(query: string) {
	/// do stuff to find things and update db...
	return liveQueryState(
          () => db.repos.where('searchWords').startsWithAnyOfIgnoreCase(query).distinct().toArray(),
	});
}
```
or lust leave `query_centre` as was and leave `things` as an observable. But I don't think this would consider good practice in svelte 5?
```typescript
/// $lib/components/things.ts
...
let search_input = $state('');
let things = $derived(query_centre.searchThings(search_input));
...
}
```
Would you say using your function would be the most idiomatic use of svelte 5?

oliverdowling:
The renaming is accidental. For me, it makes sense to have the "result" of a query and that is what I have in my project code. I see "current" being used by others and in other places, but for me it doesn't quite feel right here. When I issue a new query that has not returned yet, the value is no longer "current" but it is now an "old" value. I think "value" would probably be a good alternative, but I will continue to use "result" in my own project, so I'm happy for this to be named whatever makes the most sense for others. I tend to only use "current" for things like dynamic programming with lists/queues/graphs.

I think the reason that you are still getting flickering is because the $derived function re-runs the searchThings function that creates the query entirely, you're receiving a new state each time rather than updating the state, which is also why the dependencies are unnecessary. Try removing the “$derived” function wrapper in your first Svelte 5 example, I think that’s will be what you are looking for. I’m away and can’t test or give code examples currently.

Since liveQuery returns an observable, it is the actual resulting value from the query, and anything that subscribes to the observable will get updated. However with signals, used in Svelte 5, changes to the value itself will not cross the function boundary, hence why we need a property in the $state object, which will propagate changes.

oliverdowling:
This is just a proof of concept and can be improved, but I think this is more what you're trying to do:

```typescript
/// $lib/query_center.svelte.ts
export class QueryCenter {
	search = $state('');
	query = $state<{ value?: unknown; isLoading: boolean; error?: unknown }>({
		value: undefined,
		isLoading: true,
		error: undefined
	});
	constructor(db: Dexie) {
		$effect(() => {
			(() => [this.search])(); // no-op so it reloads
			return liveQuery(() =>
				db.repos.where('searchWords').startsWithAnyOfIgnoreCase(search).distinct().toArray()
			).subscribe(
				(value) => {
					this.query.error = undefined;
					if (value !== undefined) {
						this.query.value = value;
						this.query.isLoading = false;
					} else {
						this.query.isLoading = true;
					}
				},
				(error) => {
					this.query.error = error;
					this.query.value = undefined;
					this.query.isLoading = false;
				}
			).unsubscribe;
		});
	}
}
```

And now you can use it with less boilerplate:

```svelte
/// $lib/Component.svelte
...
const queryCenter = new QueryCenter(db);
...
<input bind:value={queryCenter.search} />
{#each queryCenter.query.value ?? [] as repo (repo.id)}
...
```

jjones315:
not sure this is actually better (i am also about 30 minutes in to using dexie), but in theory the part svelte tracks the the part live query track are different. More so hoping this triggers an idea in someone more familiar with dexie, but to me this is preferable over explicit dep tracking. would also be easy to write a wrapper for commonly used outputs like `liveQueryArray` for example, either in the adapter or userland.

so something like this would work as a primitive. 
``` ts
export class LiveQuery<T = any, TKey = IndexableType, TInsertType = T, TResult = T, TError = unknown> {
    data = $state<TResult>();
    state = $state<"error" | "loading" | "success">("loading");

    readonly isLoading = $derived(this.state === "loading");
    readonly isSuccess = $derived(this.state === "success");
    readonly isError = $derived(this.state === "error");
    error = $state<TError>();

    constructor(
        querier: () => Collection<T, TKey, TInsertType>,
        output: (query: Collection<T, TKey, TInsertType>) => PromiseExtended<TResult>
    ) {
        $effect(() => {
            const collection = querier();
            return liveQuery(() => output(collection)).subscribe(
                data => {
                    this.error = undefined;
                    if (data === undefined) {
                        this.state = "loading";
                    } else {
                        this.data = data;
                        this.state = "success";
                    }
                },
                error => {
                    this.error = error;
                    this.state = "error";
                }
            ).unsubscribe;
        });
    }
}

```
with usage like this
```ts
    let friendsQuery = new LiveQuery(
        () => db.friends.where("age").between(minAge, maxAge),
        c => c.toArray()
    );
```

oliverdowling:
This looks nice for simple synchronous queries that return a Dexie Collection. If you want to do a join, or do custom sorting/filtering so it's no longer a Collection, or have `output` be reactive (eg `someBooleanState && c.count() > 0`) then this approach will not work and you will need to declare dependencies again.

libs/dexie-svelte-query/README.md
samhoang, line 1: I'm a n00b -- I'm just testing a comment

libs/dexie-svelte-query/README.md
```diff
@@ -0,0 +1,30 @@
+# Dexie.js Svelte Query
+
+Svelte reactive state query for fetching using Dexie.js
+
```
pnwpushingpxls, line 4: Testing another comment

libs/dexie-svelte-query/src/lib/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./stateQuery.svelte.js";
```
BonaFideIT, line 1: Not sure, but this should have typescript file suffix or?

```suggestion
export * from "./stateQuery.svelte.ts";
```

dfahlander: Typescript empraces the .js suffix despite the source being ts so .js should be the correct suffix

dfahlander:
Should I merge and release this library as it is today (maybe as a 1.0.0-alpha.1)?

dfahlander:
A bit late maybe but I just merged this. Are you ok with me releasing it on npm?

dfahlander:
The library is now published on npm: https://www.npmjs.com/package/dexie-svelte-query

PR #2199, opened 2025-09-29
Migrate dexie-cloud-todo-list to vite

dfahlander:


dfahlander:
Migrated dexie-cloud-todo-list from CRA to vite

PR #2205, opened 2025-10-04
feat: useSuspendingLiveQuery

phaux:
Implemented suspending hook similar to https://github.com/dexie/Dexie.js/discussions/1276#discussioncomment-8411915

Right now it never clears the caches. I couldn't think of a simple way to implement it.

I had in mind the latest changes in Observable spec which say that observables are always hot (shared) and there is no way to detect new subscriber from observable side. This means that it's impossible to implement a BehaviorSubject which emits last value for each new subscriber. That's why I'm keeping the last value in memory for each observable. Even this is not perfect because if you pass an observable to the hook and also use it somewhere else then the first value could be missed.

I will upload an example app later which uses this so you can test it.

### DIFF
--- a/libs/dexie-react-hooks/src/useSuspendingObservable.ts
+++ b/libs/dexie-react-hooks/src/useSuspendingObservable.ts
@@ -14,36 +14,71 @@ export function useSuspendingObservable<T>(
   getObservable: () => InteropableObservable<T>,
   cacheKey: React.DependencyList
 ): T {
-  let observable = memoized(OBSERVABLE_CACHE, cacheKey, getObservable);
-  const firstValue = useSuspendingPromise(
-    () => getFirstValue(observable),
-    [observable]
-  );
-  return useObservableValue(observable, firstValue);
-}
+  let observable: InteropableObservable<T> | undefined;
+  // TODO: enable iterators in TS and remove Array.from
+  for (const [key, val] of Array.from(OBSERVABLES.entries())) {
+    if (
+      cacheKey.length === key.length &&
+      cacheKey.every((x, i) => x === key[i])
+    ) {
+      observable = val;
+      break;
+    }
+  }
+  if (!observable) {
+    observable = getObservable();
+    OBSERVABLES.set(cacheKey, observable);
+  }
 
-const OBSERVABLE_CACHE = new Map<
-  React.DependencyList,
-  InteropableObservable<any>
->();
-
-/**
- * Returns a promise that resolves with the first value emitted by the observable.
- */
-function getFirstValue<T>(observable: InteropableObservable<T>): Promise<T> {
-  return new Promise((resolve, reject) => {
-    const sub = observable.subscribe(
-      (v) => {
-        resolve(v);
-        unsub(sub);
-      },
-      (err) => {
-        reject(err);
-        unsub(sub);
+  let promise: Promise<T> | undefined = PROMISES.get(observable);
+  if (!promise) {
+    promise = new Promise<T>((resolve, reject) => {
+      if (VALUES.has(observable)) {
+        resolve(VALUES.get(observable)!);
+        return;
       }
+      const sub = observable.subscribe(
+        (val) => {
+          resolve(val);
+          VALUES.set(observable, val);
+          unsub(sub);
+          PROMISES.delete(observable);
+        },
+        (err) => {
+          reject(err);
+          unsub(sub);
+          PROMISES.delete(observable);
+        }
+      );
+    });
+    PROMISES.set(observable, promise);
+  }
+
+  use(promise);
+
+  const [value, setValue] = React.useState<T>(VALUES.get(observable));
+  const [error, setError] = React.useState<any>(null);
+
+  React.useEffect(() => {
+    const sub = observable.subscribe(
+      (val) => {
+        VALUES.set(observable, val);
+        setValue(val);
+      },
+      (err) => setError(err)
     );
-  });
+    return () => unsub(sub);
+  }, [observable]);
+
+  if (error) throw error;
+  return value;
 }
+
+const OBSERVABLES = new Map<React.DependencyList, InteropableObservable<any>>();
+
+const PROMISES = new WeakMap<InteropableObservable<any>, Promise<any>>();
+
+const VALUES = new WeakMap<InteropableObservable<any>, any>();
 
 /** Unsubscribes from an observable */
 function unsub(sub: (() => unknown) | { unsubscribe: () => unknown }) {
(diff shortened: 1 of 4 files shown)


phaux:
Example vibe coded app which uses this hook: https://stackblitz.com/~/github.com/phaux/rrdex

Notice how if you replace this hook with regular useLiveQuery then an empty/not found state will appear briefly sometimes on navigation and also entry animations will play even though they shouldn't on first render (they should only play when adding new element to a list).

If you also remove remountOnParamsChange in main then an exit animation followed by an entry animation plays for all elements in a list on every navigation, which looks cool but is not desired in this case.

I used my own list animation component in this example, but the problem also appears when using other FLIP animations libraries or React's experimental ViewTransition component.

dfahlander:
Thank you for this great piece of code! I looked through the code and it looks really thought-through, intelligent and clear. I will defenitely merge this for a new official release of dexie-react-hooks. Need some small things fixed first though:

- [x] update react in devDeps with pnpm so we get rid of some build errors of the typings
- [x] support for synchronous observables that resolve immediately
- [x] update tsconfig to support iterators and remove the Array.from() as you've commented in the code already (we shouldn't target es5 anymore! should change to ES2016)
- [x] support for React < 19 (would need to throw the promise if still pending, instead of using React.use())
- [x] support for observables and not only observable factories

phaux:
Sorry but I don't know how to run tests in this repo. I only get `SyntaxError: Unexpected identifier 'assert'` when trying to build y-dexie.

I did everything except "optimization for synchronous observables" because I don't think it's necessary? A BehaviorSubject from rxjs will already emit synchronously and the next callback will run instantly so there will be no suspense.

Now I want to add some tests but I'm gonna need help with that.

dfahlander:
I see. You need to have pnpm installed and then `cd` to the repo root and

```
pnpm install
pnpm build
pnpm test
cd libs/dexie-react-hooks
pnpm install
pnpm build 
pnpm test
```

 If we need to discuss more please join discord server https://discord.gg/huhre7MHBF

phaux:
Running tests in root fails with `Cannot find plugin "karma-webdriver-launcher".`

Building `libs/dexie-react-hooks` fails with `TS2307: Cannot find module 'y-dexie' or its corresponding type declarations.`

Building `addons/y-dexie` fails with `SyntaxError: Unexpected identifier 'assert'`

dfahlander:
> Running tests in root fails with `Cannot find plugin "karma-webdriver-launcher".`
> 
> Building `libs/dexie-react-hooks` fails with `TS2307: Cannot find module 'y-dexie' or its corresponding type declarations.`
> 
> Building `addons/y-dexie` fails with `SyntaxError: Unexpected identifier 'assert'`

Oh, y-dexie is a new dev-dependency. Need to pnpm install and build it too (addons/y-dexie). But with pnpm in the root having done `pnpm install`, `pnpm build`, `pnpm test` should work in the root.

I just cloned the repo on a fresh location to verify. Here are my commands just now that at least work on my macbook with pnpm 9.8.0 installed.

```bash
git clone https://github.com/dexie/Dexie.js dexie4
cd dexie4
pnpm install
pnpm build
pnpm test
cd addons/y-dexie
pnpm install
pnpm build
cd ../../libs/dexie-react-hooks/
pnpm install
pnpm build
pnpm test
```

phaux:
I added some tests. I simply modified the test app to use the suspending hook in one place and added additional steps and checks.

It would be nice to also check if it works with `<Suspense>` by checking if it's rendering the fallback and also with `useTransition` by checking if it's rendering with isPending equal true.

Also a check if the caches are cleared after a query is no longer in use would make the tests better.

I tried to do these but that's not so easy. Maybe I will try again to come up with something later.

dfahlander:
I added support for synchronous observables and for providing the observable itself instead of a function returning an observable. Maybe not super useful for suspense case, but in case people would prefer useSuspenseObservable() to ensure not getting undefined back for these cases, it's nice that it works equally well.

libs/dexie-react-hooks/src/useSuspendingObservable.ts
dfahlander, line 1: Thanks 👍

ryanb:
@phaux Thanks for this PR! I'm very much looking forward to using Dexie with Suspense.

I'm trying out this PR and ran into an issue. A call to `useSuspendingLiveQuery` is returning results from another query.

Looking at the code. The dependencies passed to `useSuspendingObservable` are used as the cache key. So if two different queries happen to use the same dependencies then they will use the same observable. For example.

```js
const category = useSuspendingLiveQuery(() => db.categories.get(categoryId), [categoryId])
const products = useSuspendingLiveQuery(() => db.products.where({ categoryId }).toArray(), [categoryId])
```

Here `products` could be set to `category` since they share the same dependencies.

**Update:** Looking at the example app, a string is given as the first dependency to `useSuspendingLiveQuery`, is this to handle this global cache issue? I think passing a unique string to each query will be difficult to remember and has potential to conflict unintentionally elsewhere in the code base. When not done correctly it can cause some very tricky bugs.

dfahlander:
> **Update:** Looking at the example app, a string is given as the first dependency to `useSuspendingLiveQuery`, is this to handle this global cache issue? I think passing a unique string to each query will be difficult to remember and has potential to conflict unintentionally elsewhere in the code base. When not done correctly it can cause some very tricky bugs.

Yes, that could be a real issue. Maybe the unique string could be enforced to be used as the first parameter, before the callback. That might make it harder to miss it?

dfahlander:
```ts
export const useGetCathegoryById = (categoryId: string) => useSuspendingLiveQuery(
  "useGetCathegoryById",
  () => db.categories.get(categoryId),
  [categoryId]
)
```

ryanb:
It's unfortunate there's no way to handle this internally so `useSuspendingLiveQuery` can have the same interface as `useLiveQuery`.

I played around with `useId` and `useMemo` but neither one persists on the initial re-render through Suspense.

As a "quick fix" for now I'm doing `querier.toString()` which isn't perfect since duplicate function strings don't necessarily mean the behavior is the same. However it's good enough for my needs.

phaux:
Maybe a better API would be something like

```js
useSuspendingObservable(db, "tableName", (table) => table.get(itemId), [itemId])
```

So the user is forced to give the table name as string and it will be used in the key automatically.

ryanb:
@phaux I don't know if the table name is unique enough. For example.

```js
const productsWithCategory = useSuspendingLiveQuery(db, "products", (table) => table.where({ categoryId }).toArray(), [categoryId])
const productsWithTag = useSuspendingLiveQuery(db, "products", (table) => table.where({ tagId }).toArray(), [tagId])
```

If both `categoryId` and `tagId` happen to be the same then it will share the cache.

dfahlander:
I pubilshed this version of dexie-react-hooks as dexie-react-hooks@4.2.1-alpha.1 so it's easier for anyone to try the current API out.

*NOTE: The API might change in future versions*

libs/dexie-react-hooks/src/useSuspendingObservable.ts
```diff
@@ -0,0 +1,149 @@
...
+      );
+      if (calledSynchronously) unsub(sub);
+    });
+
+    PROMISES.set(observable, promise);
+  }
+
+  if (!VALUES.has(observable)) {
+    usePromise(promise);
+  }
```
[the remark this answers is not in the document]
dfahlander: > _⚠️ Potential issue_ | _🔴 Critical_
> 
> **Call `usePromise` consistently to avoid breaking hook order.**
> 
> On React 19, `usePromise` resolves to `React.use`, so it must be called every render in the same order. The current conditional means the first render without a cached value calls the hook, but once `VALUES` is populated subsequent renders skip it, triggering React’s “rendered fewer hooks than expected” error. Please invoke `usePromise(promise)` unconditionally so the hook order stays stable while still relying on the resolved promise to short‑circuit suspense.
> 
> ```diff
> -  if (!VALUES.has(observable)) {
> -    usePromise(promise);
> -  }
> +  usePromise(promise);
> ```
> 
> 📝 Committable suggestion
> > ‼️ **IMPORTANT**
> > Carefully review the code before committing. Ensure that it accurately replaces the highlighted code, contains no missing lines, and has no issues with indentation. Thoroughly test & benchmark the code to ensure it meets the requirements.
> 
> 🧰 Tools
> 🪛 Biome (2.1.2)
> [error] 100-100: This hook is being called conditionally, but all hooks must be called in the exact same order in every component render.
> 
> For React to preserve state between calls, hooks needs to be called unconditionally and always in the same order. See https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level
> 
> (lint/correctness/useHookAtTopLevel)
> 
> 🤖 Prompt for AI Agents

Not true. React use is not a hook, it's allowed in conditional paths.

phaux:
I've been using this hook for some time and it works pretty well. I actually ended up using just this hook modified to work on top of vanilla IndexedDB+BroadcastChannel (without Dexie). I might publish it as a separate library in the future.

I was planning to push some more tests but I couldn't make them work. I can always submit what I have as a separate WIP PR if you want to merge this as is @dfahlander .

dfahlander:
> I've been using this hook for some time and it works pretty well. I actually ended up using just this hook modified to work on top of vanilla IndexedDB+BroadcastChannel (without Dexie). I might publish it as a separate library in the future.
> 
> I was planning to push some more tests but I couldn't make them work. I can always submit what I have as a separate WIP PR if you want to merge this as is @dfahlander .

I think it's fine as it is. I'll merge it to master once I get some feedback from people trying the the published alpha version of dexie-react-hooks (mainly for regression testing as we've updated tsconfig etc).

I've also laborated with an alternative API for suspense by letting a hook provide a "synchronous" API to dexie. I'll see what comes out of all this, but it's very nice that we have an npm-published version of this for people to try out!

Thanks a lot for all your effort and for a really clean and nice implementation!

libs/dexie-react-hooks/src/useSuspendingObservable.ts
```diff
@@ -0,0 +1,149 @@
...
+      incrementRef();
+
+      let calledSynchronously = false;
+      let sub: AnySubscription;
+      sub = observable.subscribe(
+        (val) => {
+          resolve(val);
+          VALUES.set(observable, val);
+          if (!sub) calledSynchronously = true;
+          else unsub(sub);
```
ryanb, line 83: @phaux @dfahlander I've noticed an issue with `useSuspendingLiveQuery` where the `querier` function is executed twice even though the dependencies and underlying data doesn't change.

Upon further investigation, calling `subscribe` on an observable appears to re-executes the query. Is this correct? I'm surprised the observable doesn't return the existing value when re-subscribing.

Since we subscribe both here and in the `useEffect` block below this is causing the `querier` to execute twice, once for each `subscribe` call.

I want to verify my assumptions here before creating an issue for this.

phaux: That seems correct. it's something that could be improved.

PR #2212, opened 2025-11-10
Dexie 5 API with chained where clauses and orderBy

dfahlander:


dfahlander:
Rebased version of this PR in #2253

Issue #10, opened 2014-05-06
Google Group settings issue


idibidiart:
<<
The group dexiejs@googlegroups.com does not allow posting through email.

Most people would use email as an interface to the mailing list  

I wonder why it's the default...

Also, if there is a way to make "receive mail for every post" as the default rather than "don't receive any emails" that would help since the default is not easily noticed and it's easy to sign up and never get any emails and assume there hasn't been any discussions.

PR #2245, opened 2026-02-13
fix(types): Allow any return type from update/modify callbacks

liz709:
## Problem

TypeScript rejects common patterns like:

```typescript
db.friends.update(id, (friend) => ++friend.age)
```

Because the callback return type was defined as `void | boolean`, but `++friend.age` returns `number`.

## Solution

Changed callback return type from `void | boolean` to `any` in:
- `Table.update()`
- `Collection.modify()`

The runtime only checks for `=== false` to cancel the modification, so any other return value is valid.

## Testing

All of these now compile:
- `db.friends.update(id, (friend) => ++friend.age)` - returns number
- `db.friends.update(id, (friend) => { friend.age++; })` - returns void  
- `db.friends.update(id, (friend) => { if (x) return false; })` - returns false to cancel
- `db.friends.toCollection().modify((friend) => ++friend.age)` - Collection.modify too

## Related

Reported in: https://github.com/dexie/dexie-web/pull/10#issuecomment-3898295972

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated callback return type specifications in Collection.modify and Table.update operations. Callbacks were previously restricted to void or boolean return types and can now return any value type. These modifications accommodate a broader range of callback implementations and usage patterns across custom modification and update operations, extending API flexibility.

### DIFF
--- a/src/public/types/collection.d.ts
+++ b/src/public/types/collection.d.ts
@@ -43,6 +43,6 @@ export interface Collection<T=any, TKey=IndexableType, TInsertType=T> {
   until(filter: (value: T) => boolean, includeStopEntry?: boolean): Collection<T, TKey, TInsertType>;
   // Mutating methods
   delete(): PromiseExtended<number>;
-  modify(changeCallback: (obj: T, ctx:{value: TInsertType}) => void | boolean): PromiseExtended<number>;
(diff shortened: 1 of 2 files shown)


liz709:
Closing this - the actual issue was in the documentation (dexie-web PR #10), not the types. The types PR author was correct from the start.

PR #2242, opened 2026-02-12
feat(blob-offloading): Client-side blob offloading and lazy resolution

liz709:
## Summary

This PR implements the client-side of blob offloading for Dexie Cloud:

### Features

1. **Blob Upload (sync time)**
   - Large blobs (≥4KB) are uploaded to blob storage before sync
   - Replaced with BlobRefs in the sync payload
   
2. **Lazy BlobRef Resolution (read time)**
   - BlobRefs from server are marked with `$unresolved` marker
   - DBCore middleware resolves BlobRefs on-demand when objects are read
   - Uses `Dexie.waitFor()` only for explicit rw transactions
   - Preserves Promise.PSD context (no async/await in middleware)

3. **Sparse Index**
   - Adds `$unresolved` index to all synced tables
   - Zero overhead for resolved objects (sparse index)

### Files Changed

- `src/sync/blobOffloading.ts` - Blob upload utilities
- `src/sync/blobResolve.ts` - BlobRef detection and download
- `src/middlewares/blobResolveMiddleware.ts` - DBCore middleware for lazy resolution
- `src/sync/applyServerChanges.ts` - Mark incoming BlobRefs as unresolved
- `src/overrideParseStoresSpec.ts` - Add $unresolved sparse index
- `src/dexie-cloud-client.ts` - Register middleware

### Testing

Tested locally with two-client sync:
- Client A creates 50KB blob → syncs to server
- Client B receives identical blob with matching checksum ✅

### Related

- Server-side PR: dexie/dexie-cloud#72
- Requires `dexie-cloud-common@1.0.58-tson.0` or later

## Summary by CodeRabbit

* **New Features**
  * Large binary data (>4KB) is offloaded to remote blob storage before sync to reduce payloads.
  * Remote blobs are resolved on read; unresolved items are marked and can be resolved in background.
  * Background eager downloader with progress reporting and a new blobMode option ('eager' or 'lazy').
  * Sync protocol bumped to v3 to support blob references.

* **Documentation**
  * Added CI publishing guide and a CI workflow for automated package publishing.

PR #2247, opened 2026-02-21
fix(dexie-cloud): cross-realm blob detection for offloading

liz709:
## Summary
Fixes blob offloading detection when running in Service Worker context where `instanceof Blob` fails due to cross-realm issues.

## Changes
- Add `getTypeTag()`, `isBlobLike()`, `isArrayBufferLike()`, `isArrayBufferViewLike()` helpers
- `shouldOffloadBlob()` now handles Service Worker realm correctly using Object.prototype.toString type tag detection
- Removed debug logging

## Testing
Tested with E2E blob-offloading tests in Service Worker context.

## New Features
- Added automatic blob offloading in Dexie Cloud sync, which efficiently stores large binary data separately, improving synchronization performance and reducing bandwidth usage.

### DIFF
--- a/addons/dexie-cloud/src/sync/blobOffloading.ts
+++ b/addons/dexie-cloud/src/sync/blobOffloading.ts
@@ -1,0 +1,307 @@
+/**
+ * Blob Offloading for Dexie Cloud
+ * 
+ * Handles uploading large blobs to blob storage before sync,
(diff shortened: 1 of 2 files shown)


liz709:
Closing in favor of PR #2242 which now includes the cross-realm blob detection fix.

PR #2250, opened 2026-02-26
Avoid unnecessary tracking of connections on Chrome 115+

francoisdoray:
Historically, Dexie tracked Dexie instances in a global "connections" array to emulate "versionchange" events and close connections on pagehide to allow pages to enter BFCache. However, Chrome natively supports "versionchange" and, since version 115, allows pages to enter the BFCache with active connections (see: http://crrev.com/c/4496128). Tracking of connections in a global "connections" array is therefore no longer necessary.

This change disables adding connections to the global "connections" array for Chrome 115+. This is important because if developers forget to call close() on a Dexie instance, it remains in the global array indefinitely. Chromium developers have identified this as a source of significant memory bloat in the browser process (hundreds of MiB in real-world cases). Not tracking connections in the global array prevents that performance issue by default.

Entering BFCache with an active IndexedDB connection is demonstrated here: https://francoisdoray.github.io/web-repros/bfcache-dexie.html

## Summary by CodeRabbit

* **New Features**
  * Enhanced database connection handling with browser-aware optimization for modern Chrome browsers.

* **Tests**
  * Added comprehensive test coverage for connection tracking and browser detection logic across different browser environments.

### DIFF
--- a/src/classes/dexie/dexie-open.ts
+++ b/src/classes/dexie/dexie-open.ts
@@ -111,7 +111,7 @@ export function dexieOpen (db: Dexie) {
           // as well as absurd upgrade version quirk for Safari.
         }
         
-        connections.push(db); // Used for emulating versionchange event on IE/Edge/Safari.
+      if (db._trackConnection) connections.push(db); // Used for BFCache support and emulating versionchange event on older browsers.
         
         idbdb.onversionchange = wrap(ev => {
             state.vcFired = true; // detect implementations that not support versionchange (IE/Edge/Safari)
--- a/src/classes/dexie/dexie.ts
+++ b/src/classes/dexie/dexie.ts
@@ -26,7 +26,7 @@ import Promise, { PSD, globalPSD } from '../../helpers/promise';
 import { extend, override, keys, hasOwn } from '../../functions/utils';
 import Events from '../../helpers/Events';
 import { maxString, connections, READONLY, READWRITE } from '../../globals/constants';
-import { getMaxKey } from '../../functions/quirks';
+import { getMaxKey, isModernChrome } from '../../functions/quirks';
 import { exceptions } from '../../errors';
 import { lowerVersionFirst } from '../version/schema-helpers';
 import { dexieOpen } from './dexie-open';
@@ -77,6 +77,7 @@ export class Dexie implements IDexie {
   _middlewares: {[StackName in keyof DexieStacks]?: Middleware<DexieStacks[StackName]>[]} = {};
   _vip?: boolean;
   _novip: Dexie;// db._novip is to escape to orig db from db.vip.
+  _trackConnection: boolean;
   core: DBCore;
 
   name: string;
@@ -112,6 +113,7 @@ export class Dexie implements IDexie {
     const {
       addons,
     } = options;
+    this._trackConnection = !isModernChrome;
     this._dbSchema = {};
     this._versions = [];
     this._storeNames = [];
--- a/src/functions/quirks.ts
+++ b/src/functions/quirks.ts
@@ -19,3 +19,18 @@ export let getMaxKey = (IdbKeyRange: typeof IDBKeyRange) => {
     return maxString;
   }
 }
+
+export const isModernChromeInternal = (userAgent: string) => {
+  const chromeMatch = userAgent.match(/Chrome\/(\d+)/);
+  return chromeMatch && parseInt(chromeMatch[1], 10) >= 115;
+}
+
+/** 
+ * Chrome 115 allows a page to enter the BFCache with active connections
+ * and transactions, as long as they're not blocking others. Additionally,
+ * it supports the "versionchange" event natively. As a result, there
+ * is no need to track Dexie connections in a global array to support
+ * BFCache or "versionchange" on modern Chrome.
+ */
+export const isModernChrome = typeof navigator !== 'undefined' &&
+  isModernChromeInternal(navigator.userAgent);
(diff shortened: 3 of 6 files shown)


francoisdoray:
Extra details:

- I'm a Chromium contributor (fdoray@chromium.org). We've observed a large number of live IndexedDB connections on some sites using Dexie, which leads to significant memory bloat in the browser process.
- You can see the browser-side leak by navigating to [this repro page](https://francoisdoray.github.io/web-repros/dexie-leak.html) and clicking the button to start leaking. If you monitor chrome://histograms/IndexedDB.NumConnections.OnCreateAbove10k in a new tab, you'll see samples confirming that there is a large number of live IndexedDB connections, despite the Dexie objects seeming to be eligible for GC (but they aren't because they're retained by the connections array in Dexie). 
- Pages can now enter Chrome's BFCache with open IndexedDB connections. If a BFCache'd page holds a connection with a version that doesn't match the version requested by another page, it is evicted from the BFCache (closing the connection on pagehide would have avoided that, but we believe that preventing the leak is more important to the user than preventing BFCache eviction on this ~rare event). If a BFCache'd page holds a transaction that conflicts with a transaction being opened by another page, it is evicted from BFCache (closing a connection is not an effective protection against that eviction, since ongoing transactions are not immediately closed on connection close).
- I'm happy to answer any extra questions you might have!

dfahlander:
It's true that `Dexie.connections` holds all Dexie instances that have an open IndexedDB connection. The array is static and hinders garbage collection. On the other hand, indexedDB connections are resources that need to be closed explicitely, just like file handles, WebSockets etc. 

Even though not documented on dexie.org, nor used by any known addons, the `Dexie.connections` has been an API surface for a very long time and changing its behavior might break code in theory.

### Questions

* How does Chromium handle leaking non-closed [IDBDatabase](https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase) instances? Does it auto-close them on garbage collection? If not, there would still be an issue for apps not closing Dexie instances properly.

* If Chromium DOES auto-close raw IDBDatabase instances, is this a generic behavior for resources or specific for IDBDatabase?

* If we'd do the same with Dexie.connections using FinalizationRegistry, would that be a viable solution or do you see any performance downsides of doing so?

* Since `Dexie.connections` is an API surface that has been there for at least 8 years (even though not documented on dexie.org), I'd like to explore what options we could have, see below:

### Options to dismantelling Dexie.connections immediately

1. Use FinalizationRegistry to auto-remove from Dexie.connections on GC. This would align with Chromium's auto-close of IDBDatabases (if that's the case). Deprecate Dexie.connections in next major
2. Have a default `maxConnections` in Dexie constructor options, set to 20, 100 or similar and warn + recycle old connections when exceeded.

### Other possible actions
* Update dexie docs to more clearly advice to 1) Avoid creating Dexie instances dynamically (for performance reasons) and  2) If doing that anyway, state clearly that instances need to be explicitely closed.
* Implement `[Symbol.dispose]` in Dexie and update dexie docs to using the `using` keyword when creating Dexie instances dynamically.

francoisdoray:
###  Questions -> Answers

1. On GC, Chromium collects and closes non-referenced `IDBDatabase` instance. For example, this connection would be collected and closed on a GC, even tough it isn't closed explicitly:

```
const request = indexedDB.open("DB", 1);
request.onerror = (event) => {};
request.onsuccess = (event) => {
  const db = event.target.result;
  const transaction = db.transaction("store", "readonly");
  const store = transaction.objectStore("store");
  const readRequest = store.get(1); 
  readRequest.onsuccess = () => { };
  readRequest.onerror = (event) => { };
};
```

If Javascript code keeps a reference to the `IDBDatabase`, like Dexie does with the `connections` array, Chromium cannot collect it.

2. Chromium only collects and closes unreferenced `IDBDatabase` instances. The `GPUDevice` obtained from `adapter.requestDevice()` is similar.

3. A `FinalizationRegistry` would be perfect. To keep the overhead minimal, there could still be an option to not populate the `connections` array for users who don't need it.

4. Ok, we can explore other options!

###  Options to dismantelling Dexie.connections immediately

From Chromium performance point of view, using a `FinalizationRegistry` (to allow collecting+closing connections that aren't externally referenced) **and** adding a `maxConnections` argument would be ideal (to ensure that Javascript code doesn't create egregious amounts of connections and keeping references to them unknowingly - unrelated to Dexie, we see Web pages holding 10k+ connections).

### Other possible actions

Updating the Dexie docs to more clearly state that Dexie objects must be explicitly closed is a nice bonus. When an IDBDatabase object is explicitly closed, resources can be reclaimed immediately, instead of waiting for the next GC cycle, which is ideal for performance. However, I wouldn't rely only on that to fix the problem (we should also have the `FinalizationRegistry` + `maxConnections`), as experience tells us that developers tend to forget `close()` calls, even when it's documented.

I'm not so sure that `[Symbol.dispose]` is practical, given that `Dexie` objects are used across multiple scopes.



Would you like me to send pull requests to use a `FinalizationRegistry` and add a `maxConnections` parameter?

dfahlander:
Thanks for your detailed response!

Proposed solution then would be:

* Use FinalizationRegistry solution to auto-evict Dexie instances from Dexie.connections.

* New options argument to Dexie constructor `trackConnections: boolean | number` that could either false to skip tracking connections, true to do it without a limit (as current behaviour) or a number of the limit we keep in Dexie.connections. Default 10. If num connections are exceeded, we would just recycle the Dexie.connections array - not forcibly close them.

If you have a chance to update or file new PR it would be much apprechiated and I would prioritize its review and merging. If not, let me know and I can try fix it later.

Thanks again,
David

dfahlander:
Superceded by #2254

PR #2252, opened 2026-03-01
perf: Use getAllRecords() for reverse queries (2-5x faster)

### LINKED ISSUE #2183: new getAllRecords() API
Not really an issue but browsers are going add a new `.getAllRecords` API https://github.com/w3c/IndexedDB/pull/461. Might be useful for Dexie.

liz709:
## Summary

Implements support for the new IndexedDB `getAllRecords()` API ([Interop 2026](https://webkit.org/blog/17818/announcing-interop-2026/#getallrecords-for-indexeddb)) for reverse iterations. This provides **2-5x faster reads**, especially for reverse queries.

## Motivation

As discussed in #2183, browsers are now implementing `getAllRecords()` which allows fetching records with direction, eliminating the need for cursor iteration on reverse queries.

- Chrome: ✅ Supported since v115
- Firefox: 🔜 Coming in 2026 (Interop 2026)
- Safari: 🔜 Coming in 2026 (Interop 2026)

### Benchmark

From [Microsoft's demo](https://microsoftedge.github.io/Demos/idb-getallrecords/):
- **2x faster** for normal queries
- **5x faster** for reverse queries

## Changes

- **`src/functions/quirks.ts`**: Add `hasGetAllRecords` feature detection
- **`src/public/types/dbcore.d.ts`**: Add `reverse?: boolean` to `DBCoreQueryRequest`
- **`src/public/types/global.d.ts`**: Add type declarations for `getAllRecords()` API
- **`src/dbcore/dbcore-indexeddb.ts`**: Use `getAllRecords({ direction: 'prev' })` for reverse queries
- **`src/classes/collection/collection.ts`**: Update `toArray()` and `primaryKeys()` to pass `reverse` flag
- **`test/tests-getallrecords.js`**: Comprehensive tests for reverse queries

## Affected Methods

These methods now benefit from the optimization when called with `.reverse()`:

- `collection.reverse().toArray()`
- `collection.reverse().primaryKeys()`
- `table.orderBy(index).reverse().toArray()`
- `table.reverse().limit(n).toArray()`

## Backward Compatibility

When `getAllRecords()` is not available (older browsers), the code falls back to cursor iteration - exactly as before.

## Testing

- ✅ Build passes
- ✅ New tests added for reverse queries
- ✅ Fallback behavior preserved

Closes #2183

## Summary by CodeRabbit

* **New Features**
  * Direction-aware data retrieval using modern IndexedDB getAll/getAllKeys/getAllRecords when supported.
  * Runtime detection of modern IndexedDB features to choose optimized query paths.
  * Public typings extended to support getAll options, record shapes, and an optional query direction.

* **Behavior**
  * Queries, limit(), reverse() and live-query optimistic updates consistently respect query direction.

* **Tests**
  * Added tests covering reverse(), limit(), orderBy(), and direction-aware retrieval.

### DIFF
--- a/src/dbcore/dbcore-indexeddb.ts
+++ b/src/dbcore/dbcore-indexeddb.ts
@@ -292,7 +292,9 @@ export function createDBCore (
       return (request: DBCoreQueryRequest) => {
         return new Promise<DBCoreQueryResponse>((resolve, reject) => {
           resolve = wrap(resolve);
-          const {trans, values, limit, query, direction} = request;
+          const {trans, values, limit, query} = request;
+          // Normalize direction once - undefined means 'next'
+          const direction = request.direction ?? 'next';
           const nonInfinitLimit = limit === Infinity ? undefined : limit;
           const {index, range} = query;
           const store = (trans as IDBTransaction).objectStore(tableName);
@@ -306,14 +308,14 @@ export function createDBCore (
             const options = {
               query: idbKeyRange,
               count: nonInfinitLimit,
-              direction: direction || 'next'
+              direction
             };
             const req = values ?
                 (source as any).getAll(options) :
                 (source as any).getAllKeys(options);
             req.onsuccess = event => resolve({result: event.target.result});
             req.onerror = eventRejectHandler(reject);
-          } else if (hasGetAll && (!direction || direction === 'next')) {
+          } else if (hasGetAll && direction === 'next') {
             const req = values ?
                 (source as any).getAll(idbKeyRange, nonInfinitLimit) :
                 (source as any).getAllKeys(idbKeyRange, nonInfinitLimit);


dfahlander:
I think using getAll with the new overloaded `options` argument would be more accurate for the query endpoint , rather than getAllRecords since we only return values and not keys. That new overload is also new in indexeddb 3.

https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex/getAll

liz709:
Good catch! ✅ Updated to use `getAll({ query, count, direction: 'prev' })` and `getAllKeys({ query, count, direction: 'prev' })` instead of `getAllRecords()`.

This is more accurate since:
- `toArray()` only needs values → `getAll(options)`
- `primaryKeys()` only needs keys → `getAllKeys(options)`

We keep `getAllRecords` only for feature detection (to know if the browser supports IDB 3.0).

nickchomey:
nitpick: getAllRecords is still mentioned in many places (especially in the tests), despite it having been removed in favour of getAll.

dfahlander:
> nitpick: getAllRecords is still mentioned in many places (especially in the tests), despite it having been removed in favour of getAll.

Thanks. Renamed test just now in master branch

PR #2267, opened 2026-03-19
fix(cloud): resolve BlobRefs in WebSocket-pushed modify/update operations

liz709:
## Problem
When server pushes changes via WebSocket containing BlobRefs in `modify`/`update` operations, the client stores them in IDB without the `_hasBlobRefs` marker. The `blobResolveMiddleware` never triggers resolution, so the app sees raw BlobRef objects (`{_bt, ref, size}`) instead of the actual values.

## Root Cause
`applyServerChanges()` only called `markIfHasBlobRefs()` for `insert`/`upsert` operations. `modify` and `update` operations were missing the check entirely.

## Solution
Added `markChangeSpecIfHasBlobRefs()` that inspects changeSpec values for BlobRefs and injects `_hasBlobRefs: 1` into the changeSpec. When Dexie merges the changeSpec into the existing IDB object, the marker gets set, and `blobResolveMiddleware` picks it up on the next read.

Applied to both `modify` (single changeSpec) and `update` (array of changeSpecs).

## Changes
- `applyServerChanges.ts`: Added `markChangeSpecIfHasBlobRefs()` + calls in modify/update cases

## Summary by CodeRabbit

* **Documentation**
  * Updated `maxStringLength` option documentation to explicitly specify valid configuration bounds and constraints.

* **Improvements**
  * Enhanced validation for the `maxStringLength` configuration option, now enforcing values within the required range of 100 to 32,768 to prevent invalid configurations.
  * Improved handling of blob references during cloud synchronization to ensure proper data merging during update operations.

### DIFF
--- a/addons/dexie-cloud/src/sync/applyServerChanges.ts
+++ b/addons/dexie-cloud/src/sync/applyServerChanges.ts
@@ -20,6 +20,32 @@ function markIfHasBlobRefs(obj: unknown): void {
   }
 }
 
+/**
+ * For changeSpecs (partial updates from modify/update operations),
+ * check if any value in the spec contains BlobRefs. If so, add
+ * _hasBlobRefs = 1 to the changeSpec so it gets merged into the
+ * existing object in IDB.
+ */
+function markChangeSpecIfHasBlobRefs(changeSpec: {[key: string]: any}): void {
+  for (const key of Object.keys(changeSpec)) {
+    const value = changeSpec[key];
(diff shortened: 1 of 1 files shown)


liz709:
Closing — this was the wrong approach. The _hasBlobRefs marker is set by the sending client, not the receiving one. The actual fix is PR #2268 which makes cache-middleware invalidate instead of optimistically applying BlobRef values.

PR #2297, opened 2026-04-16
fix(webkit): retry tempTransaction on 'Cannot inject key into script value' UnknownError

liz709:
## Problem

Safari/WebKit throws `UnknownError: Cannot inject key into script value` on certain iOS versions when Dexie performs `put()`/`add()` operations. This appears to be a transient WebKit timing issue triggered on cold start or after wake-from-background — the WebKit IndexedDB engine is not yet fully initialized.

- Reported by user in Dexie Discord #general
- GitHub issue: https://github.com/dexie/Dexie.js/issues/2296
- Only affects Safari/WebKit — Chrome on iOS is not affected

## Fix

In `tempTransaction()`, catch `UnknownError` with message containing `'inject key'` and retry the transaction with exponential backoff (100ms base, 1.5x multiplier, max 20 retries). The DB is **not** reopened — just the transaction is retried.

```
Retry schedule: 100ms, 150ms, 225ms, 337ms, ... (up to 20 retries)
Max total wait: ~30 seconds before giving up
```

A `console.warn` is emitted on the first retry to aid debugging.

## Testing

Needs testing on affected iOS Safari version. @markfilius has offered to test a canary build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction reliability in WebKit browsers (Safari) by automatically retrying transient “inject key” UnknownError occurrences.
  * Retries employ exponential backoff with a capped number of attempts to avoid infinite loops.
  * Emits a one-time console warning on the first retry; if retries are exhausted the original error is surfaced.

### DIFF
--- a/src/classes/dexie/transaction-helpers.ts
+++ b/src/classes/dexie/transaction-helpers.ts
@@ -145,9 +145,9 @@ export function enterTransactionScope(
         return trans._completion.then(() => x);
       })
       .catch((e) => {
-        trans._reject(e); // Yes, above then-handler were maybe not called because of an unhandled rejection in scopeFunc!
         // Workaround for WebKit "Cannot inject key into script value" UnknownError.
         // Only retry root transactions (not sub-transactions) since sub-transactions share parent's IDBTransaction.
+        // Don't reject the transaction if we're going to retry — let the retry handle it.
         // https://github.com/dexie/Dexie.js/issues/2296
         if (
           !parentTransaction &&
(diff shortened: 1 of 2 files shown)


src/classes/dexie/transaction-helpers.ts
```diff
@@ -130,6 +146,34 @@ export function enterTransactionScope(
       })
       .catch((e) => {
         trans._reject(e); // Yes, above then-handler were maybe not called because of an unhandled rejection in scopeFunc!
```
dfahlander: Don't do trans._reject() if we're retrying on this!

PR #2310, opened 2026-06-09
fix(live-query): clone results on cache read when cache: 'cloned' is used (issue #2309)

### LINKED ISSUE #2309: useLiveQuery stops updating React component state in Dexie 4.4.x
I have a live simulation function that updates the db.headToHeadMatches store every 100 - 200 ms, pretty intensive, but it works well in Dexie 4.3.

```
await db.headToHeadMatches4.put(updatedMatch);

```

```
  const match = useLiveQuery(() => {
      return db.headToHeadMatches
        .where("[teamId1+teamId2+year]")
        .equals([teamId1, teamId2, year])
        .last();
        // the query does run every update.
  }, [teamId1, teamId2, year]);
// match <- but this match is not updated...
```
However, in 4.4.0, I have to either return a new object to make it work or clone the object before the put action:

```
const match = useLiveQuery(async () => {
    const h = await db.headToHeadMatches4
      .where("[teamId1+teamId2+year]")
      .equals([teamId1, teamId2, year])
      .last();
    return h ? { ...h } : null; 
}, [teamId1, teamId2, year]);
```
or
```
await db.headToHeadMatches4.put({ ...updatedMatch });

```

Would you please tell me if it is an expected behavior in Dexie 4.4? The query inside useLiveQuery does fetch the new data on every update, it is just that the useLiveQuery never returns an updated match.

liz709:
Closes #2309. This PR ensures that query results retrieved from the query cache are deeply cloned on every read if the 'cloned' cache option is used. This fixes a regression where React hooks like useLiveQuery did not trigger a re-render when mutating a returned object in-place and putting it back to the database, as well as protecting the cache against direct modifications by the user.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live-query caching to better control how optimistic results are prepared and returned, preventing unintended mutations when result freezing is disabled.

* **Tests**
  * Updated and modernized the live-query test suite, including refactored helpers, a more structured results/mutation matrix runner with reliable coordination and cleanup, and refreshed regression coverage for optimistic updates, transactions, error cases, and cache/mutation isolation behaviors.

### DIFF
--- a/src/live-query/cache/cache-middleware.ts
+++ b/src/live-query/cache/cache-middleware.ts
@@ -336,15 +336,16 @@ export const cacheMiddleware: Middleware<DBCore> = {
             }
             subscribeToCacheEntry(cacheEntry, container!, requery, signal);
             return cacheEntry.promise.then((res: DBCoreQueryResponse) => {
+              const result = applyOptimisticOps(
+                res.result,
+                req,
+                tblCache?.optimisticOps,
+                downTable,
+                cacheEntry!,
+                freezeResults
+              ) as any[]; // readonly any[]
               return {
-                result: applyOptimisticOps(
-                  res.result,
-                  req,
-                  tblCache?.optimisticOps,
-                  downTable,
-                  cacheEntry!,
-                  freezeResults
-                ) as any[], // readonly any[]
+                result: freezeResults ? result : deepClone(result),
               };
             });
           },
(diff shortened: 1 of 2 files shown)


dfahlander (approved):
Me and Liz had a deep conversation around this. I found that the PR would degrade performance due to duplicate deepClone() calls. My idea was to skip the deepClone() around line 299 in cache-middleware.ts since we anyway do a deepClone() again after awaiting the promise.

Liz concluded this was a very good idea and committed a new commit that removed deepClone() from line 299.

I asked Liz to think deeply around other possible scenarios that we might have missed by removing the extra deepClone(). Liz thought it through thorougly and was 100% certain this was not of any issue since the data we clone isn't modified anywhere and never forwarded to user so it can't possibly be manipulated in any way.

I agree with Liz conclusion.

PR #2311, opened 2026-06-11
Add missing table4 parameter typing to transaction method

renbaoshuo:
## Summary by CodeRabbit

* **New Features**
  * Expanded transaction API to support additional table parameters, allowing more complex multi-table operations.

### DIFF
--- a/src/public/types/dexie.d.ts
+++ b/src/public/types/dexie.d.ts
@@ -101,6 +101,7 @@ export interface Dexie {
     table: string | Table,
     table2: string | Table,
     table3: string | Table,
+    table4: string | Table,
     table5: string | Table,
     scope: (trans: TXWithTables<this>) => PromiseLike<U> | U
   ): PromiseExtended<U>;


dfahlander:
That part in dexie.d.ts is generated from src/public/types/dexie.d.ts.

Jumping over table4 in one of the overloaded methods should be fixed there.

But regarding the limited number of tables: It is always possible to use an array argument instead and then there is no limitation of number of tables:

```ts
await db.transaction('rw', [...tables], callback);
```

dfahlander:
But I see now that the code was actually changing the correct file. Merging.

dfahlander:
Thanks!

renbaoshuo:
By reading the source code, I understand that arrays can be used. However, the presence of a false "table5" remains confusing unless fixed.

Thank you for your explanation, and thank you so much for this fantastic project!

dfahlander:
> By reading the source code, I understand that arrays can be used. However, the presence of a false "table5" remains confusing unless fixed.
> 
> Thank you for your explanation, and thank you so much for this fantastic project!

Agree with it fixed a confusing typo!! Thanks back!

PR #2325, opened 2026-07-31
fix(dbcore): don't crash when an index is named 'constructor'

### LINKED ISSUE #1920: constructor as table key makes diexie crash
I have created https://github.com/AriaMoradi/vitejs-vite-3goktz to demonstrate the issue.

tarann26:
## What

Declaring a table whose primary key or an index is named `constructor` crashes `db.open()`:

```js
db.version(1).stores({ items: 'id, constructor' });
await db.open(); // TypeError: indexList.push is not a function
```

## Why

In `src/dbcore/virtual-index-middleware.ts`, `indexLookup` is a plain object literal used as a string-keyed map:

```ts
const indexLookup: { [indexAlias: string]: VirtualIndex[] } = {};
// ...
const indexList = (indexLookup[keyPathAlias] = indexLookup[keyPathAlias] || []);
indexList.push(virtualIndex);
```

When an index's key-path alias is `constructor`, `indexLookup['constructor']` resolves through `Object.prototype` to the inherited `Object` constructor (truthy, non-array), so the `|| []` fallback never runs and `.push()` is called on a function. This affects any index name that collides with an `Object.prototype` member (`constructor`, `toString`, `valueOf`, `hasOwnProperty`, …).

## Fix

Give `indexLookup` no prototype, so a bracket lookup can only ever resolve to an own key:

```ts
const indexLookup: { [indexAlias: string]: VirtualIndex[] } =
  Object.create(null);
```

All four `indexLookup` accesses are plain bracket get/set, so this is safe with no API change.

## Test

Adds a regression test in `test/tests-misc.js` that opens a db with an index named `constructor`, inserts a row, and reads it back via `where('constructor').equals(...)` — failing before the change with `TypeError: indexList.push is not a function` and passing after.

Fixes #1920.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of database indexes named `constructor`, preventing errors during database operations.
  * Strengthened dictionary-based index lookups to avoid conflicts with inherited object properties.

### DIFF
--- a/src/dbcore/virtual-index-middleware.ts
+++ b/src/dbcore/virtual-index-middleware.ts
@@ -46,7 +46,8 @@ export function createVirtualIndexMiddleware(down: DBCore): DBCore {
     table(tableName: string) {
       const table = down.table(tableName);
       const { schema } = table;
-      const indexLookup: { [indexAlias: string]: VirtualIndex[] } = {};
+      const indexLookup: { [indexAlias: string]: VirtualIndex[] } =
+        Object.create(null);
       const allVirtualIndexes: VirtualIndex[] = [];
 
       function addVirtualIndexes(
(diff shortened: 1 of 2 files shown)


dfahlander:
Thanks! <3

PR #2328, opened 2026-08-28
fix(dexie-export-import): don't emit a trailing comma when a chunk is fully filtered out

### LINKED ISSUE #1070: Filtered Export produces invalid JSON file with an extra comma in data row array

buildwithakhila:
Fixes #1070.

`exportDB()`'s chunked-write loop decided whether to prepend a comma based on whether the *previous* chunk had written rows. With an active `filter`, a chunk can come back with zero matching rows — the old code still queued a comma for it, so if a match was followed only by non-matching chunks through the end of the table, the export ended with a comma before the closing `]` (invalid JSON), exactly as reported.

Fix: track whether *any* row has been written yet with a running flag, and only add the separating comma (and only write a slice) once a chunk is confirmed, after filtering, to actually contain rows.

Added a regression test in `edge-cases.ts` following the pattern of the existing `numRowsPerChunk` tests (10 rows, `numRowsPerChunk: 3`, filter matching only the first row — guarantees a later chunk is filtered out entirely). It fails with the old code, reproducing the exact `SyntaxError` from the issue, and passes with the fix. Ran the full addon test suite locally — all 7 tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed database exports that could produce invalid JSON when filtered chunks contained no matching rows.
  * Ensured exported data includes only matching records without trailing separators.

* **Tests**
  * Added regression coverage for filtered, chunked exports to verify valid JSON output and correct record counts.


dfahlander:
Thanks <3

